PR #28, opened 2016-01-20
Update french translation

M2ck:
Update About string


xsoh:
@M2ck  thank you so much! could you please merge the two commits into one commit so it doesn't make the history with two commits.
Thanks :D

M2ck:
@xsoh I have no idea how do that, sorry my skills on GitHub are limited (using browser only).
I will send you a new PR and I close this one.

PR #39, opened 2016-05-01
buidl with circleCI

zoff99:
added commands to build with circleCI
and screenshot will be created every time to check that the app is running

https://circleci.com/gh/zoff99/Etar-Calendar/31#artifacts

https://circle-artifacts.com/gh/zoff99/Etar-Calendar/31/artifacts/0/tmp/circle-artifacts.p8u6982/capture_app_running.png

capture_app_running

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -84,6 +84,8 @@ dependencies:
       ls -al ;
       cp -av app-signed-aligned.apk $CIRCLE_ARTIFACTS/etar_circleci_$CIRCLE_SHA1.apk || exit 1
     - qrencode -o $CIRCLE_ARTIFACTS/QR_apk.png 'https://circle-artifacts.com/gh/zoff99/Etar-Calendar/'${CIRCLE_BUILD_NUM}'/artifacts/'${CIRCLE_NODE_INDEX}'/tmp/'`basename $CIRCLE_ARTIFACTS`'/'"etar_circleci_$CIRCLE_SHA1.apk" ; exit 0
+    - qrencode -o $CIRCLE_ARTIFACTS/QR_artifacts.png 'https://circle-artifacts.com/gh/zoff99/Etar-Calendar/'${CIRCLE_BUILD_NUM}'#artifacts' ; exit 0
+
 test:
   pre:
     - mksdcard -l e 180M sdcard.img
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,135 @@
+machine:
+  timezone:
+    Europe/Vienna
+  environment:
+    ZEMU_: "appemu23"
+    EMU_: "appemu21"
+    XEMU_: "appemu10"
+    Zsdpath: "/sdcard"
+    sdpath: "/storage/sdcard"
+    Xsdpath: "/mnt/sdcard"
+    ANDROID_HOME: /usr/local/android-sdk-linux
+    ANDROID_SDK: /usr/local/android-sdk-linux/
+    _SDK_: /usr/local/android-sdk-linux/
+    GRADLE_OPTS: '-Dorg.gradle.jvmargs="-Xmx1000m -XX:+HeapDumpOnOutOfMemoryError"'
+    GRADLE_HOME: ~/g/gradle-2.10
+    PATH: ~/g/gradle-2.10/bin:$PATH
+general:
+  artifacts:
+    #-./app/build/outputs/logs/
+dependencies:
+  cache_directories:
+    - ~/.android
+    - ~/.gradle
+  override:
+    - echo "dummy"
+  pre:
+    - sudo apt-get update > /dev/null 2> /dev/null
+    - sudo apt-get install xvfb > /dev/null 2> /dev/null
+    - sudo apt-get install xdotool > /dev/null 2> /dev/null
+    - sudo apt-get install telnet > /dev/null 2> /dev/null
+    - sudo apt-get install x11-utils > /dev/null 2> /dev/null
+    - sudo apt-get install xvkbd > /dev/null 2> /dev/null
+
+    - gradle -v
+    - cd ~ ; wget https://services.gradle.org/distributions/gradle-2.10-bin.zip
+    - mkdir ~/g ; cd ~/g ; unzip ~/gradle-2.10-bin.zip
+    - gradle -v
+
+    - echo y | android update sdk --no-ui --all --filter "tools"
+    #- android list sdk --all | grep -i tools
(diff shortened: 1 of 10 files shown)


jspricke:
Thanks for the pull request, I will put some comments inline, but in general I'm ok with it.

circle.yml
```diff
@@ -0,0 +1,135 @@
...
+    Xsdpath: "/mnt/sdcard"
+    ANDROID_HOME: /usr/local/android-sdk-linux
+    ANDROID_SDK: /usr/local/android-sdk-linux/
+    _SDK_: /usr/local/android-sdk-linux/
+    GRADLE_OPTS: '-Dorg.gradle.jvmargs="-Xmx1000m -XX:+HeapDumpOnOutOfMemoryError"'
+    GRADLE_HOME: ~/g/gradle-2.10
+    PATH: ~/g/gradle-2.10/bin:$PATH
+general:
+  artifacts:
+    #-./app/build/outputs/logs/
```
jspricke: Why is this commented? If it's useful, why not uncomment it, otherwise I would propose to remove it.

circle.yml
```diff
@@ -0,0 +1,135 @@
...
+dependencies:
+  cache_directories:
+    - ~/.android
+    - ~/.gradle
+  override:
+    - echo "dummy"
```
jspricke: Is this really needed?

circle.yml
```diff
@@ -0,0 +1,135 @@
...
+  pre:
+    - sudo apt-get update > /dev/null 2> /dev/null
+    - sudo apt-get install xvfb > /dev/null 2> /dev/null
+    - sudo apt-get install xdotool > /dev/null 2> /dev/null
+    - sudo apt-get install telnet > /dev/null 2> /dev/null
+    - sudo apt-get install x11-utils > /dev/null 2> /dev/null
+    - sudo apt-get install xvkbd > /dev/null 2> /dev/null
+
```
jspricke: I would propose to remove the redirects, so we get an error log, in case there is one, also do we need anything of it besideds xdotool?

circle.yml
```diff
@@ -0,0 +1,135 @@
...
+    - echo y | android update sdk --no-ui --all --filter platform-tools > /dev/null 2> /dev/null
+
+    - echo y | android update sdk --no-ui --all --filter extra-android-m2repository > /dev/null 2> /dev/null
+    - echo y | android update sdk --no-ui --all --filter extra-google-m2repository > /dev/null 2> /dev/null
+    - echo y | android update sdk --no-ui --all --filter extra-android-support > /dev/null 2> /dev/null
+
+    - echo y | android update sdk --no-ui --all --filter platform-tools-preview > /dev/null 2> /dev/null
+    - echo y | android update sdk --no-ui --all --filter sys-img-armeabi-v7a-android-23 > /dev/null 2> /dev/null
+    - echo y | android update sdk --no-ui --all --filter sys-img-armeabi-v7a-addon-google_apis-google-23 > /dev/null 2> /dev/null
+    - echo y | android update sdk --no-ui --all --filter addon-google_apis-google-23 > /dev/null 2> /dev/null
```
jspricke: I would propose to remove the output redirects here as well

circle.yml
zoff99: yes

circle.yml
zoff99: yes you need to first "update" and also "telnet"
but the others do not hurt

circle.yml
```diff
@@ -0,0 +1,135 @@
...
+
+    - zip -d app/build/outputs/apk/app-debug-unaligned.apk META-INF/\*     # remove signature !!
+    - cp -av app/build/outputs/apk/app-debug-unaligned.apk ~/app.apk ;
+      cd ~/ ;
+      echo xxxxxxrm -f ~/.android/debug.keystore ;
+      ls -al ~/.android/debug.keystore ;
+      if [ ! -f ~/.android/debug.keystore ]; then echo "*** generating new signer key ***" ;
+      echo "*** generating new signer key ***";
+      echo "*** generating new signer key ***";
+      keytool -genkey -v -keystore ~/.android/debug.keystore -storepass android -keyalg RSA -keysize 2048 -validity 10000 -alias androiddebugkey -keypass android -dname "CN=Android Debug,O=Android,C=US" ;
```
jspricke: What's the use of signing every apk with a new debug key? I would propose to simply provide the Etar-Calendar-debug-unaligned.apk.

circle.yml
zoff99: sure, if you want to

circle.yml
zoff99: remove it, but you also need to remove the "artifact:" lines.
but i would not change anything at all on the 1st merge.
i tested it like this.

circle.yml
zoff99: you can remove the redirects. but if there is any error it will turn "RED" and stop anyway

circle.yml
zoff99: it does NOT !
circle CI will save it between runs. a new key will only be created on the first run.
hence the "if [ ! -f ~/.android/debug.keystore ]"

echo xxxxxxrm -f ~/.android/debug.keystore ; --> this line is only there to quickly make a new key if needed.

but its your call. i run many apps with this logic :-)

zoff99:
in general this is just a templare that works. you can change to your liking.
do you have a circleCI account? its free and if you merge this PR all your commits will automatically be built and tested. also all your PRs will.

circle.yml
jspricke: Huh, why don't you want to change it? The usual way is to either give hints why something should stay in it's current way or comply with the reviewer and change it. Btw., changing is easy, just push a new commit to the branch.

circle.yml
jspricke: huh, where do we need telnet? And "do not hurt" is not an argument, so remove them ;).

zoff99:
@xsoh are you going to merge this one?

zoff99:
ok. you won't merge.

xsoh:
@zoff99 Actually, I really don't know anything about circleCI.
If it is working very well then I have no comments on it and we can go with it. 

Thanks @zoff99 👍  !

zoff99:
thanks.

it should be fine. you can merge it.
if it should fail (which i dont expect) i will fix the commands for circle.

if you merge, you should login to https://circleci.com with your github account just once,
to activate building all commits there from that point on.

zoff99:
@xsoh do you still want to merge this? it's now been a long time ...

PR #52, opened 2016-06-08
Solves #1

Gitsaibot:
Should solve #1

### DIFF
--- a/src/com/android/calendar/CalendarToolbarHandler.java
+++ b/src/com/android/calendar/CalendarToolbarHandler.java
@@ -141,13 +141,6 @@ public class CalendarToolbarHandler {
             dayOfWeek = DateUtils.formatDateRange(mContext, mFormatter, mMilliTime, mMilliTime,
                     DateUtils.FORMAT_SHOW_WEEKDAY, mTimeZone).toString();
         }
-        String[] split = dayOfWeek.toLowerCase().split(" ");
-        StringBuilder sb = new StringBuilder();
-        for (String word : split) {
-            sb.append(word.substring(0, 1).toUpperCase() + word.substring(1));
-            sb.append(" ");
-        }
-        dayOfWeek = sb.toString();
         return dayOfWeek;
     }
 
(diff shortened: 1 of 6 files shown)


src/com/android/calendar/CalendarToolbarHandler.java
```diff
@@ -141,7 +141,14 @@ private String buildDayOfWeek() {
             dayOfWeek = DateUtils.formatDateRange(mContext, mFormatter, mMilliTime, mMilliTime,
                     DateUtils.FORMAT_SHOW_WEEKDAY, mTimeZone).toString();
         }
-        return dayOfWeek.toUpperCase();
```
jspricke: Shouldn't it be enough to remove the `.toUpperCase()` here?
Can you change your pull request accordingly?

Gitsaibot: I tried this first but it didn't work. There is a problem with the addition Heute, Morgen, Gestern. Maybe you find a better or easier way to achieve this ;). But until now I didn't find another way...

jspricke: I tried it over here and it worked fine. What problem did you get exactly?

Gitsaibot: Also mal schnell auf Deutsch ist einfacher für mich ;). Wenn ich nur .toUpperCase() entferne hab ich bei mir das Problem das er das hier anzeigt: **GESTERN**, Dienstag. Bei den Tagen ohne Zusatz passt es soweit.

jspricke: (I would propose to stay with English, so others can follow).

For me it shows "Heute, Sonntag" tried Android 4.4.4 and 5.1

Gitsaibot: In the Android emulator(Nexus 5X, API 23) i get this: 
device-2016-06-12-132915

On my Phone and in german it shows the same as for you. I don't if it's the language !?

jspricke: I guess it's a language thing. Still, my proposal would be to only remove the `.toUpperCase()` as your solution is not portable among languages. Could you change your pull request accordingly?

Gitsaibot: Okay I can change that but what will we do with the english language ?

jspricke:
No need to close this, just force push into your branch ;)

Issue #50, opened 2016-06-03
No separate column for week number


Gitsaibot:
I think it is better to integrate the Week number in the Monday colum(maybe top left corner). This saves space. Besides the number should be a little bit smaller.

PanderMusubi:
Note that first day of the week can differ from locale to locale or calendar details.

PR #55, opened 2016-06-13
Solves #50

Gitsaibot:
Hmm I didn't understand git fully but actually "Solves 8" should be a separate Pull request :(.

### DIFF
--- a/res/values/colors.xml
+++ b/res/values/colors.xml
@@ -64,8 +64,8 @@
     <color name="month_day_number">#FF555555</color>
     <color name="month_mini_day_number">#FF000000</color>
     <color name="month_day_number_other">#FF999999</color>
-    <color name="month_week_num_color">#FF999999</color>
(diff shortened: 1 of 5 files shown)


jspricke:
Thanks for the pull request. Can you please move the different features into separate requests, as you noted? You need to use separate branches for that. You can do something like that to fix it:
```
git checkout master
git branch issue8
git rebase -i origin/master
```
no you should be able to select only the right commits. Also, you can add 70374d6 into the first commit by replacing pick with fixup. While at it, can you add a proper commit log (i.e. what it changes and some details)? Simply change pick to reword. Thanks!

jspricke:
Once you are fine with your branch, you can do a `git push -f` to update this pull request

jspricke:
No need to close this, just force push the fixed commits ;)

jspricke:
Hi @Gitsaibot I fixed your patches and merged them. Hope you are fine with that. Thanks a lot for the contribution!

Issue #56, opened 2016-06-13
Change Color from "Add Button" 


Gitsaibot:
I think it looks better when Action Bar and Add Button have the same color. Or is there a reason for the current behavior ?

PanderMusubi:
Please change the bar of the widget into the same color. Preferably use a color picker that offers colors (including grays) from Material Design but allows override with RGB value.

smichel17:
If you want to be fully material, your accent color should be more saturated than your primary color. Eg, #1dc1ab

jspricke:
@smichel17 this is closed already, but it would be great if you could have a look at #89 and comment what to improve there. Thanks!

PR #60, opened 2016-06-19
Solves #56

Gitsaibot:


jspricke:
Can you please adopt this as described in #55 as well? Thanks!

PR #91, opened 2016-10-01
Correct select behavior 

### LINKED ISSUE #72: Wrong shared checkable behavior in "CalendarView".xml
Wrong behavior, more then one is selected:
device-2016-07-14-222147

It should be like that, only one should be selectable: 
device-2016-07-14-222030
#89

Gitsaibot:
Correct wrong select behavior. Now only one entry is selectable at the same time.


jspricke:
Hi @Gitsaibot thanks for separating this. Could you please add a proper commit message? You can find some documentation on it here: http://chris.beams.io/posts/git-commit/. Thanks!

Issue #102, opened 2016-11-10
Calendars to display menu broken


davidhedlund:
Reproduce
* Click on Menu -> Calendars to display
* Click on the hardware menu button to get options (touch button left to Home push button on most Samsung devices).

This will crash Etar:
> Unfortunately, Etar has stopped.

Gitsaibot:
I can confirm this. It's a NPE.

davidhedlund:
[mylabel:bug](https://github.com/search?utf8=%E2%9C%93&q=commenter%3Adavidhedlund+%22mylabel%3Abug%22+in%3Acomments&type=Issues&ref=searchresults)

davidhedlund:
@Gitsaibot Thanks.

davidhedlund:
@Gitsaibot What is NPE?

Gitsaibot:
Null Pointer Exception

davidhedlund:
@Gitsaibot Thanks.

PR #103, opened 2016-11-10
Prevent NPE in Search and Select Calendar Activity. Solves #102

Gitsaibot:
Prevents crashes in Search and Select Calendar Activity.


jspricke:
Thanks! Btw. if you write "Fixes" instead of "Solves" the corresponding issue should be closed automatically: https://github.com/blog/1386-closing-issues-via-commit-messages

PR #105, opened 2016-11-13
 Add an option to choose days per Week.

### LINKED ISSUE #78: Display less days on the screen
Hi,

Nobody have a large screen and thus, display seven days on a small screen looks really cluttered. 

Is it possible to have a choice to display 3/4 days? This would be more handy.

Bests!

Gitsaibot:
Add an option to choose days per Week, Translate Goto in German, Show which Theme is selected under Settings. Fixes #78

### DIFF
--- a/res/values-de/strings.xml
+++ b/res/values-de/strings.xml
@@ -274,7 +274,10 @@
     <item quantity="other" msgid="8779598489398048259">"<xliff:g id="EVENT_COUNT">%d</xliff:g>-mal"</item>
   </plurals>
     <string name="acessibility_recurrence_choose_end_date_description" msgid="5899029840837836711">"Enddatum ändern"</string>
-  <string name="preferences_theme">Darstellung</string>
-  <string name="preferences_dark_theme">Dunkel</string>
-  <string name="preferences_light_theme">Hell</string>
+  <string name="preferences_theme">"Darstellung"</string>
+  <string name="preferences_dark_theme">"Dunkel"</string>
+  <string name="preferences_light_theme">"Hell"</string>
+  <string name="preferences_days_per_week_dialog">"Anzahl Tage pro Woche"</string>
+  <string name="preferences_days_per_week_title">"Anzahl Tage pro Woche"</string>
+  <string name="goto_date">"Gehe zu..."</string>
 </resources>
--- a/src/com/android/calendar/GeneralPreferences.java
+++ b/src/com/android/calendar/GeneralPreferences.java
@@ -58,10 +58,12 @@ import ws.xsoh.etar.R;
 public class GeneralPreferences extends PreferenceFragment implements
         OnSharedPreferenceChangeListener, OnPreferenceChangeListener, OnTimeZoneSetListener {
     // Preference keys
+    public static final String KEY_THEME_PREF = "pref_theme";
     public static final String KEY_HIDE_DECLINED = "preferences_hide_declined";
     public static final String KEY_WEEK_START_DAY = "preferences_week_start_day";
     public static final String KEY_SHOW_WEEK_NUM = "preferences_show_week_num";
     public static final String KEY_DAYS_PER_WEEK = "preferences_days_per_week";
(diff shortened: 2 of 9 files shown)


src/com/android/calendar/GeneralPreferences.java
```diff
@@ -58,10 +58,12 @@
 public class GeneralPreferences extends PreferenceFragment implements
         OnSharedPreferenceChangeListener, OnPreferenceChangeListener, OnTimeZoneSetListener {
     // Preference keys
+    public static final String KEY_THEME_PREF = "pref_theme";
```
jspricke: can you put this (and the related changes into an extra commit?

src/com/android/calendar/Utils.java
```diff
@@ -682,7 +682,12 @@ public static boolean getHideDeclinedEvents(Context context) {
 
     public static int getDaysPerWeek(Context context) {
         final SharedPreferences prefs = GeneralPreferences.getSharedPreferences(context);
-        return prefs.getInt(GeneralPreferences.KEY_DAYS_PER_WEEK, 7);
+        return Integer.valueOf(prefs.getString(GeneralPreferences.KEY_DAYS_PER_WEEK, "7"));
+    }
+
+    public static int getMDaysPerWeek(Context context) {
+        final SharedPreferences prefs = GeneralPreferences.getSharedPreferences(context);
+        return Integer.valueOf(prefs.getString(GeneralPreferences.KEY_MDAYS_PER_WEEK, "7"));
```
jspricke: This looks like we get the same functionality twice. Would it make sense to merge this, or add a second configuration option?

res/values-de/strings.xml
```diff
@@ -274,7 +274,10 @@
     <item quantity="other" msgid="8779598489398048259">"<xliff:g id="EVENT_COUNT">%d</xliff:g>-mal"</item>
   </plurals>
     <string name="acessibility_recurrence_choose_end_date_description" msgid="5899029840837836711">"Enddatum ändern"</string>
-  <string name="preferences_theme">Darstellung</string>
-  <string name="preferences_dark_theme">Dunkel</string>
-  <string name="preferences_light_theme">Hell</string>
+  <string name="preferences_theme">"Darstellung"</string>
```
Gitsaibot: I added the quotation marks because in the translation editor all other words had them.

src/com/android/calendar/month/MonthByWeekFragment.java
```diff
@@ -353,7 +353,7 @@ protected void setUpHeader() {
         }
         return loader;
     }
-
+
```
Gitsaibot: Yes I squashed an then it looks like this :(. First delete and than add. The line is the same.

src/com/android/calendar/Utils.java
Gitsaibot: There was a premature function to change the weekdays in month view. I didn't want to delete this so I only renamed it because I thought DaysPerWeek is better for Week view.  MDaysPerWeek I used for the month view if we need this sometime.

jspricke: good idea, so I would propose to add a configuration option for it as well. Could you do that?

Gitsaibot: Yes, but I'm unsure if this function make really sense on month view. I tried this out and it's nothing I really need. If someone comes up with an use-case we can still add this !?

res/values-de/strings.xml
Gitsaibot: Hm, I can undo this but then this 3 words are the only one without quotation marks. In my opinion it's better to have one pattern. But your are right this has nothing to do with this commit. Nevertheless I translate "Goto" too and 3 words more or less makes no difference or ?

jspricke: the normal git advise is to do only one thing in one commit. Never mind, this is fine for me.

src/com/android/calendar/Utils.java
jspricke: ok, let's leave it for later.:)

PR #106, opened 2016-11-23
Add gradle wrapper

daktak:
user can compile using gradlew script if gradle not present


jspricke:
I'm against this, there are many ways to install gradle.

PR #109, opened 2016-12-11
Update gradle file to enable "Instant Run". Also update Build and Sdk…

Gitsaibot:
… dependencies.


jspricke:
I've changed mavenCentral to jcenter, because the gradle plugin is not on the former and pushed it to master.

PR #115, opened 2016-12-13
Fix the issue of Calendar

Gitsaibot:
Set the time for t1 but not for t.

CRs-fixed: 530201
Change-Id: I9d6cebf11e3e42672c07466dbb737464fbf18daf


jspricke:
HI @Gitsaibot could you preserve the author when picking commits from other branches? Doing git cherry-pick <SHA1> should work.

jspricke:
git is smart enough to handle that, as I just tried and pushed it ;).

PR #116, opened 2016-12-17
prepareForSave: Dont set time in allday events

Gitsaibot:


jspricke:
Is this a required change for #117? Would you be ok, if I simply cherry-pick the commit from cm?

jspricke:
Done :)

PR #131, opened 2017-01-02
Fix sound setting on Android 6 by not de-registering settings listeners

a1291762:
Because choosing a sound is a separate activity on Android 6, the
onPause and onStop methods are called so the app doesn't observe
the settings values changing.

There doesn't seem to be a good reason to de-register from settings
listeners. I couldn't find any API notes or reference code that
indicated this should ever be done.


jspricke:
Thanks!

PR #132, opened 2017-01-04
Support sharing calendars via vcs and add ability to import/export from sdcard

Gitsaibot:


src/com/android/calendar/icalendar/IcalendarUtils.java
```diff
@@ -16,27 +16,39 @@
...
-import android.util.Log;
 import com.android.calendar.CalendarEventModel;
 
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
+import java.io.*;
```
jspricke: Can we move this back to individual imports?

res/values/cm_strings.xml
```diff
@@ -60,4 +61,14 @@
     <!-- Label values for the default start page preference. -->
     <string name="default_start_title">Default view</string>
     <string name="default_start_last">Previously used view</string>
+
+    <!-- Strings for buttons in drop down menu -->
+    <string name="buttons_list_0">Day</string>
```
jspricke: Do we need them somewhere?

src/com/android/calendar/AllInOneActivity.java
```diff
@@ -895,9 +895,6 @@ public void onDateSet(DatePickerDialog dialog, int year, int monthOfYear, int da
             return true;
         } else if (itemId == R.id.action_search) {
             return false;
-        } else if (itemId == R.id.action_delete_events) {
```
jspricke: Can you comment why this is not needed anymore?

res/values/cm_strings.xml
```diff
@@ -60,4 +61,21 @@
...
```
jspricke: Do we need them somewhere?

src/com/android/calendar/icalendar/IcalendarUtils.java
```diff
@@ -16,24 +16,43 @@
...
 import android.provider.CalendarContract;
-import android.util.Log;
+import android.system.ErrnoException;
+import android.system.OsConstants;
 import com.android.calendar.CalendarEventModel;
 
-import java.io.File;
-import java.io.FileOutputStream;
-import java.io.IOException;
+import java.io.*;
```
jspricke: Can we move them back to individual imports?

src/com/android/calendar/icalendar/IcalendarUtils.java
```diff
@@ -16,24 +16,43 @@
...
-import java.util.Date;
+import java.util.Random;
 import java.util.TimeZone;
 
 /**
  * Helper functions to help adhere to the iCalendar format.
  */
 public class IcalendarUtils {
 
-    private static int sPermittedLineLength = 75; // Line length mandated by iCalendar format
```
jspricke: Is this needed?

src/com/android/calendar/icalendar/IcalendarUtils.java
```diff
@@ -189,4 +308,53 @@ public static long convertTimeToUtc(long millis, String localTimeZone) {
...
+        InputStream in = null;
+        OutputStream out = null;
+        try {
+            in = new FileInputStream(src);
+            out = new FileOutputStream(dst);
+
+            byte[] buf = new byte[1024];
+
+            try {
+                for (int len; (len = in.read(buf)) > 0; ) {
```
jspricke: I would make this a while loop.

src/com/android/calendar/icalendar/IcalendarUtils.java
Gitsaibot: The spec says: At the most basic level, iCalendar files are organized into individual lines of text (called content lines) which are limited to 75 octets in length.

src/com/android/calendar/icalendar/IcalendarUtils.java
Gitsaibot: Why ?

res/values/cm_strings.xml
Gitsaibot: No. But they where in the original commit. I can delete them

src/com/android/calendar/AllInOneActivity.java
Gitsaibot: This was added with commit "Add ability to import/export from sdcard" but we don't have a all delete events function so I removed it.

src/com/android/calendar/icalendar/IcalendarUtils.java
jspricke: Sorry, I as asking about why making the variable public.

src/com/android/calendar/icalendar/IcalendarUtils.java
jspricke: Because that's what it is semantically. But I could live with this as well.

src/com/android/calendar/AllInOneActivity.java
jspricke: Would it make sense to add the delete feature first?

Gitsaibot: I was not sure if we need this because we have already a delete function and the question is if we need this really ? Besides it's an big commit and a lot of things to merge/change...

src/com/android/calendar/icalendar/IcalendarUtils.java
Gitsaibot: I would let it as it is. It makes mergin commits easier ;).

Sur3:
How do i export the calender to sd card? i dont find the option in the menu?!

Sur3:
Ohh ok, is that feature already planed then?

zoff99:
you can already export/import calenders and events with:
https://f-droid.org/repository/browse/?fdid=org.sufficientlysecure.ical
https://f-droid.org/repository/browse/?fdid=org.dgtale.icsimport

or create offline calenders:
https://f-droid.org/repository/browse/?fdid=org.sufficientlysecure.localcalendar&fdpage=15&fdstyle=grid

try if those suit your needs? dont bloat every singel app on android if possible

Sur3:
@zoff99 Yes I'm using your second link right now, but starting an other app exporting the calender and transferring it to my other devices to import it every time i made new events is kind of annoying, it would be nice if the .ics file is automatically updated when an event is changed, exporting the whole calender always takes a long time and i think only the calender itself knows what events are changed and therefore could update such a file efficiently.
I'm using orage on my pc as calender which automatically stores inside an .ics file and can show different other .ics calender files too, so one just has to share the .ics files, no additional import/export, thats pretty practical, i'm kind of missing such a feature in android calender apps.. :-/

Sur3:
I asked in Calender Import/Export if an .ics calender backend would be possible, probably that would be the best solution.. https://github.com/SufficientlySecure/calendar-import-export/issues/67

PR #134, opened 2017-01-06
Process RSVP Intent Without Event Editor

Gitsaibot:


jspricke:
Could you keep the original imports and use them in the changes? That would simplify the diff a lot ;).

jspricke:
Ok, fine with me. Could you resolve the conflict so I can merge this? Thank!

jspricke:
Sorry to say, but no. While this would work, it would result into a merge of master into RSVP into master, leaving a confusing history. As a rule of thumb, never merge the master into something else, only the other way round. Could you rebase the commits onto the new master instead?

jspricke:
Note that you can force push to your branch and reopen the pull request. That way it's easier to follow, as the discussion is right there.

jspricke:
Yeah, that's kind of counter intuitive, you have to reopen it before
pushing. If you did so already, you could get the old sha1 from the
reflog, push that one, repoen and then do the force push.
But for this one a new pull request is fine as well.

Cheers Jochen

PR #135, opened 2017-01-07
Improve "Map handling"

Gitsaibot:


jspricke:
Could you add a commit to move the `onReceive` back to it's old place? That would simplify the diff.

PR #136, opened 2017-01-08
Improve alarm handling

Gitsaibot:
Support doze mode and setExact on newer devices

### DIFF
--- a/src/com/android/calendar/Utils.java
+++ b/src/com/android/calendar/Utils.java
@@ -200,9 +200,7 @@ public class Utils {
      * Returns whether the SDK is the KeyLimePie release or later.
      */
     public static boolean isKeyLimePieOrLater() {
-        // TODO when SDK is set to 19, switch back to this:
-//      return Build.VERSION.SDK_INT > Build.VERSION_CODES.JELLY_BEAN_MR2;
-        return "KeyLimePie".equals(Build.VERSION.CODENAME);
+      return Build.VERSION.SDK_INT > Build.VERSION_CODES.JELLY_BEAN_MR2;
     }
 
     public static int getViewTypeFromIntentAndSharedPref(Activity activity) {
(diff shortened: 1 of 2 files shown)


src/com/android/calendar/Utils.java
```diff
@@ -196,6 +196,15 @@ public static boolean isJellybeanOrLater() {
       return Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN;
     }
 
+    /**
+     * Returns whether the SDK is the KeyLimePie release or later.
+     */
+    public static boolean isKeyLimePieOrLater() {
+        // TODO when SDK is set to 19, switch back to this:
```
jspricke: Isn't it set to 19 already?

src/com/android/calendar/Utils.java
```diff
@@ -196,6 +196,13 @@ public static boolean isJellybeanOrLater() {
       return Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN;
     }
 
+    /**
+     * Returns whether the SDK is the KeyLimePie release or later.
+     */
+    public static boolean isKeyLimePieOrLater() {
+      return Build.VERSION.SDK_INT > Build.VERSION_CODES.JELLY_BEAN_MR2;
```
jspricke: I don't have a device to test, but wouldn't this return true all the time?

Gitsaibot: I thought Build.VERSION.SDK_INT shows the SDK Version from your phone and if you don't use KitKat or later it returns false. I can delete it but then the alarm for older phones before KitKat won't work. I know we have API 19 so it's your decision ;).

jspricke: As you see, we have API 19, so I expect more stuff to brake if someone would try. So +1 for removing it.

PR #140, opened 2017-01-14
Improve "Map handling"

Gitsaibot:


res/values-zh-rCN/cm_strings.xml
```diff
@@ -20,4 +21,22 @@
     <string name="snooze_delay_dialog_title">设定暂停提醒的重复时间</string>
     <string name="preferences_default_snooze_delay_title">默认暂停提醒时长</string>
     <string name="preferences_default_snooze_delay_dialog">默认暂停提醒时长</string>
+    <string name="no_map">手机里没有找到地图</string>
+
+    <!-- Delete events -->
+    <string name="events_delete">删除事件</string>
```
jspricke: These are not needed again, right?

Issue #128, opened 2016-12-28
datetimepicker does not honor vibrate settings


ddast:
The dialogs for selecting the date and time in the activity to edit an event and in the "Go to..." dialog do always vibrate even when "Vibrate on tap" is disabled in the system settings.  This is a bug in the library [com.android.datetimepicker](https://github.com/xsoh/android_frameworks_opt_datetimepicker/tree/a62b3c24ea7bedc32de00567e430a706a6ec135c).

Fortunately this has been fixed in the CyanogenMod repository from which this library is forked from (it should also be fixed in your fork but not in the version included in this project).  Reading the commit log, [this](https://github.com/CyanogenMod/android_frameworks_opt_datetimepicker/commit/3fc32c45f5efc4ce4b91cbcdd925d9b30f67046e) should be the commit fixing this problem.  I compiled Etar with the [current version](https://github.com/CyanogenMod/android_frameworks_opt_datetimepicker/tree/66a7685e2e9a524989f840bd13217beeb3fc6342) (66a7685 at the time of writing) of datetimepicker and checked that the problem is indeed fixed.

So it would be great if you could update datetimepicker to fix this.

Note that the current version of datetimepicker requires a small change in Etar to work.  The version you use defines the color `@color/white`but this was later [removed](https://github.com/CyanogenMod/android_frameworks_opt_datetimepicker/commit/a09b3c940e98b8606a16a94b48b6d0121d9d3635) since it is already included in android as `@android:color/white`.  However, you use `@color/white` in res/values/styles.xml and res/values/themes.xml, so these should also be replaced with `@android:color/white`.  Apart from that I did not encounter any problems with the new version.

Gitsaibot:
It would be nice if you could do a Pull Request since you know all necessary changes. I think this is the fastest way to get this in master ;).

ddast:
Thanks for your answer.  I will gladly create a pull request for the changes regarding the white color.  However, I don't know how or if I can create a pull request that updates datetimepicker.

jspricke:
@ddast thanks for the pull request and sorry for the delay. You can send the other pull reuest here: https://github.com/xsoh/android_frameworks_opt_datetimepicker

ddast:
I tried, but the current upstream version of datetimepicker is in branch cm-14.1 which does not exist in the fork you linked to and it seems I can only create pull requests for existing branches and not new branches.

jspricke:
[...]

PR #143, opened 2017-01-17
Update submodule datetimepicker

ddast:
This updates datetimepicker to the newest version available in your [fork](https://github.com/xsoh/android_frameworks_opt_datetimepicker) (branch cm-13.0).
Although there is a more current version in the [upstream repo](https://github.com/CyanogenMod/android_frameworks_opt_datetimepicker) by CyanogenMod, the problem described in #128 is already fixed by this commit.

### DIFF
--- a/res/layout/all_in_one_material.xml
+++ b/res/layout/all_in_one_material.xml
@@ -29,8 +29,7 @@
                 style="@style/Widget.CalendarAppTheme.ActionBar"
                 android:layout_width="match_parent"
                 android:layout_height="?attr/actionBarSize"
-                android:minHeight="?attr/actionBarSize"
-                app:layout_scrollFlags="scroll|enterAlways" />
+                android:minHeight="?attr/actionBarSize" />
         </android.support.design.widget.AppBarLayout>
 
         <android.support.design.widget.FloatingActionButton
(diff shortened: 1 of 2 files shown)


jspricke:
I'm not sure about this. The original upstream repo is here: https://android.googlesource.com/platform/frameworks/opt/datetimepicker/. It would be great to follow it as close as possible and only use forks if needed. Could you check if that version would work as well?

ddast:
Etar compiles and works with the upstream version by Google.  There is only one minor problem I found:  The "Done" button was not fully visible on my test phone.  CyanogenMod fixed this by this [one-liner](https://github.com/CyanogenMod/android_frameworks_opt_datetimepicker/commit/c6eafb2076b66fff6d658d4743d7b0c522133e1b).

The other changes by CyanogenMod (there are only few) do not look relevant to me.

jspricke:
Thanks a lot! I would propose to move the datetimepicker to the Google master branch and cherry pick the one commit.
Also, I would propose to keep it in sync with Google and rebase our patches. This would mean force pushes to the master, but I don't expect many changes there anyhow. @xsoh @Gitsaibot would be great to get your input here as well.

jspricke:
I've created an alternative pull request in #148. Comments welcome.

jspricke:
@ddast please push the commit to a separate branch and open a new pull request ;).

ddast:
Yeah sorry.  I saw my mistake too late.  I guess I can close this anyway.

PR #148, opened 2017-01-24
Update submodules

jspricke:
* new module chips (moved from ex)
* update calendar, colorpicker and ex to latest current master
* update datetimepicker and timezonepicker to latest rebase branch
* remove unused assets from timezonepicker


jspricke:
@ddast @Gitsaibot would be great if you could have a look at this. Thanks!

ddast:
I tested this pull request and did not encounter any problems.  The vibrate bug in datetimepicker is gone as already discussed.  For the other libraries I did not notice any changes, but I don't know all places where they are used by heart so I could be missing something.

For what it's worth:  I think moving to the upstream Google repos is the right way to go.

PR #166, opened 2017-02-13
Fix height in month view

### LINKED ISSUE #158: Something changed the Box hight in Month view
If I have 3 appointments on a day the month view show only 1 and in the left corner a "+2". I have to tap on the day to see the other appointments. 

With Version 1.0.8
box_hight
With Current master
box_hight1

@ddast It comes from your commit #151. Removing app:layout_scrollFlags="scroll|enterAlways" changes the Box hight. Can you have a look into that ?

ddast:
Fixes #158.  I hope I didn't miss any side effects this time since the month view is quite complex.

The height of a box in the month view is calculated [here](https://github.com/Etar-Group/Etar-Calendar/blob/master/src/com/android/calendar/month/MonthByWeekAdapter.java#L264) using the formula `(parent.getHeight() + (parent.getTop() * 2)) / (mNumWeeks + 1)`.  The origin of this formula is not clear to me but apparently it sets the height such that _approximately_ `mNumWeeks` boxes are visible.
 
After commit 3ebc9c11fafb3baf212731ec3739ce4ad4c9e583 the toolbar is no longer scrollable and this reduces the result of `parent.getHeight()` leading to the behavior described in #158.  I think the reason is that with a scrollable toolbar the area occupied by the toolbar is included in `parent.getHeight()` since the toolbar could be removed by scrolling.

However, after making the toolbar unscrollable `parent.getHeight()` should give the actual size excluding the area occupied by the toolbar.  Thus, the correct height of the boxes can be calculated by the simple formula `parent.getHeight() / mNumWeeks` as done in this pull request.

This does not restore the exact original behavior but the height of the boxes is now calculated such that `mNumWeeks=6` boxes (and a tiny bit more if you look carefully) are visible.  In the original behavior approximately `mNumWeeks=6` boxes were visible (afaik it can be either more or less depending on the resolution). For comparison see the following screenshots.

Original behavior
1_scrollable_toolbar

After making the toolbar unscrollable
2_noscrollable_toolbar

Unscrollable toolbar + this pull request
3_fixheight

### DIFF
--- a/src/com/android/calendar/month/MonthByWeekAdapter.java
+++ b/src/com/android/calendar/month/MonthByWeekAdapter.java
@@ -261,8 +261,7 @@ public class MonthByWeekAdapter extends SimpleWeeksAdapter {
             selectedDay = mSelectedDay.weekDay;
         }
 
-        drawingParams.put(SimpleWeekView.VIEW_PARAMS_HEIGHT,
-                (parent.getHeight() + (parent.getTop() * 2)) / (mNumWeeks + 1));
+        drawingParams.put(SimpleWeekView.VIEW_PARAMS_HEIGHT, parent.getHeight() / mNumWeeks);
         drawingParams.put(SimpleWeekView.VIEW_PARAMS_SELECTED_DAY, selectedDay);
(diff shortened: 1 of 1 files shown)


jspricke:
Looking good, thx!
@Gitsaibot any comments before I merge?

Issue #168, opened 2017-02-14
Whole-day event gets created one day too short


sharst:
When creating a whole-day event from e.g. Monday till Thursday, the event gets inserted as only from Monday until Wednesday, i.e. always one day too short.
The same happens when editing an event

Steps to reproduce:

- Create new event (all-day, from Monday till Thursday)
- Event gets added on Monday, Tuesday and Wednesday
- Edit the event, save it without changes
- The event is now on Monday and Tuesday only.

Tested on 1.0.8

Gitsaibot:
Did you really test this with version 1.0.8 ? I can't reproduce this there.

sharst:
Hm, weird, I just tried again and the issue is still the same, and the version (Settings --> About) is definitely 1.0.8.

jspricke:
I gave @sharst a test version of the new 1.0.9, so it's not really 1.0.8 (my fault). I can reproduce it with 1.0.9 over here as well. @Gitsaibot can you confirm? I will try to bisect it today.

jspricke:
The commit in questions is af31beddede05a1a340f0183de83f83e5c283ab4, coming from #116.

jspricke:
So we are not the first running into this, see original commit here: 29190975b9238dd6841f822f82f2fb83b0557f36 and https://github.com/SufficientlySecure/calendar-import-export/issues/50

Gitsaibot:
Yes it's #116. @jspricke do you create a pull request ?

jspricke:
I would like to find out what this commit should actually fix first. I exported an all day event with and without af31beddede05a1a340f0183de83f83e5c283ab4 but they where the same. @Gitsaibot do you remember the context why you proposed it?

jspricke:
This is fixed for now. Thanks for the report @sharst.

PR #173, opened 2017-02-20
Revert "prepareForSave: Dont set time in allday events"

jspricke:
This reverts commit af31beddede05a1a340f0183de83f83e5c283ab4.


jspricke:
I think it's more important to fix #168 and do a new release fast. But it would still be interesting to find out what this is about.

PR #197, opened 2017-03-20
Allow user to squeeze display so that whole day is visible, amd omit …

rparkins999:
Allow user to squeeze display so that whole day is visible, amd omit some hour numbers if they get too close together.

I want to be able to see all my events in the week view, so that I can see if I have any commitments on a particular day. This is viable on devices with high resolution displays, provided you only show every n'th hour number if they get too close together. This patch saves the last cell height set by gesture in the preference, and then adjusts it for each day or week if there are fewer all day events at the top.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.2.3'
+        classpath 'com.android.tools.build:gradle:2.3.0'
     }
 }
 
--- a/src/com/android/calendar/DayView.java
+++ b/src/com/android/calendar/DayView.java
@@ -2469,14 +2469,15 @@ public class DayView extends View implements View.OnCreateContextMenuListener,
     private void drawHours(Rect r, Canvas canvas, Paint p) {
         setupHourTextPaint(p);
 
-        int y = mCellHeight + mHoursTextHeight / 2 + HOUR_GAP;
         int totCellHeight =  mCellHeight + HOUR_GAP;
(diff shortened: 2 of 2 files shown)


rparkins999:
Thanks for the prompt reply. I don't think that github allows multiple pull requests. If you look at a pull request, you see what is in my fork when you look, not what was there when I created the pull request. So if I upload some changes, do a pull request, upload some more changes, and then do another pull request, you can only see both sets of changes together. At least that is how it seems to work for me.

I could in theory back out and do two separate commits. However the gradle build changes are forced on me by keeping my copy of Android Studio up to date, so I'm not sure if I can back those out and still build. I know I can't back out the platform updates, because I couldn't bulld the unmodified git clone without them. In any case the code changes are all in one file (DayView.java), so it should be easy enough to pick out just that file.

After a bit of experimenting, I can reproduce the glitch that you illustrate, so I'll look into it. I think that the problem is associated with redrawing after a squeeze and recalculating the layout not being properly synchronised. If so, it may require significant rearrangement of the code to get rid of it.

Incidentally, is there some reason why you set the typeface style for event descriptions to bold hidden inside getEventLayout(), rather than more obviously in the initialisation of mEventTextPaint? I didn't like the bold style so I've removed it in my copy [I haven't pushed that change], but I'm looking at setting a user preference for EVENT_TEXT_FONT_SIZE, since a smaller size works on my high resolution screen.

rparkins999:
I think patch 1798a19 should fix the fast squeeze glitch

rparkins999:
Closed by mistake

jspricke:
> I don't think that github allows multiple pull requests.

Use a new branch with a fitting name for every pull request.

PR #225, opened 2017-05-30
swap position of delete and save event actions

### LINKED ISSUE #210: Swap 'export to sd card' and 'delete'
When viewing an event, I keep hitting export when I meant to delete, because I expect delete to be the rightmost action; would you consider swapping the positions of those menu items?

smichel17:
Fixes #210


smichel17:
Note: I couldn't decide whether edit should also be to the right of save. Ended up just going with this because I seldom misclick edit, but am happy to redo this with the other positioning if you think that would be better.

PR #268, opened 2017-11-01
Updated de (German) translation

Luensche:
Commit made via Stringlate


Luensche:
It's also rearranged as the English original is arranged

jspricke:
What would be the benefit of arranging both files equally? Also, I don't think the English one is a good example for such a file. Could you please split the rearranging into a separate commit?

Issue #220, opened 2017-05-19
Support RTL


xsoh:
Support right-to-left UI is important to languages like Arabic, Persian, Hebrew..etc.

smichel17:
Which parts currently do not support RTL?

xsoh:
@smichel17 currently, **Week view** and **Month view** are not working properly. this because  [SimpleWeekView.java](https://github.com/Etar-Group/Etar-Calendar/blob/master/src/com/android/calendar/month/SimpleWeekView.java) doesn't recognize nor draw on the correct direction when RTL language is set.

elig0n:
RTL looks to be supported in Month view (on latest version 1.0.26 at least) but not on Day view nor Week View where the hours and/or weekdays remain on the left. 

Yet Is there a reason why RTL is not enabled so far ?

Personally I think that even with a reverse week view/hours-column the app is better off with RTL enabled since most of the app is already translated and right now it only leads to confusion and disorientation.

PR #301, opened 2018-02-01
 Improve dark theme on event_info dialog

Gitsaibot:


res/layout/contact_item.xml
```diff
@@ -18,14 +18,14 @@
     xmlns:android="http://schemas.android.com/apk/res/android"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
-    android:layout_marginLeft="7dip"
-    android:paddingLeft="4dip"
+    android:layout_marginStart="7dip"
```
xsoh: Using `*Start` and `*End` instead of `*Left` and `*Right` will make the app initially supporting RTL, which is good 😄 , however, the content view of where the days are shown will be the opposite side. because [SimpleWeekView.java](https://github.com/Etar-Group/Etar-Calendar/blob/master/src/com/android/calendar/month/SimpleWeekView.java) doesn't support RTL. This will make a confusion to the user. That's why I recommend not do it until fixing (Support RTL #220).

Gitsaibot: That was my thought ;). And whilst android:supportsRtl="true" is not set I thought it should be no problem but I can revert it if you want. I don't use left to right anyway ;)

Gitsaibot: @xsoh Should I remove RTL or not ?

PR #313, opened 2018-04-11
Updated the Norwegian translation

DandelionSprout:
I added some extra strings from the English values to the Norwegian strings.xml to better serve Etar's app settings, and created a store description translation.

### DIFF
--- a/res/values-nb/strings.xml
+++ b/res/values-nb/strings.xml
@@ -176,7 +176,7 @@
     <string name="menu_general_preferences" msgid="2812705323334260708">"Generelle innstillinger"</string>
     <string name="menu_about_preferences" msgid="6124375438237674973">"Om Kalender"</string>
     <string name="preferences_title" msgid="8045855493262131773">"Innstillinger"</string>
-    <string name="preferences_general_title" msgid="2069075513760573117">"Kalender visningsinnstilling"</string>
+    <string name="preferences_general_title" msgid="2069075513760573117">"Kalender-visningsinnstillinger"</string>
     <string name="preferences_reminder_title" msgid="5199606534191180129">"Varsler og påminnelser"</string>
     <string name="preferences_hide_declined_title" msgid="1849295943669890034">"Skjul avslåtte aktiviteter"</string>
     <string name="preferences_week_start_day_title" msgid="4619489296444901622">"Uke starter på"</string>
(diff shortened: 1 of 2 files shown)


DandelionSprout:
I've attempted to improve the translation now. If there's any strings that are still missing, then I'll have to be told about which ones I missed.

PS: I couldn't find any strings about snooze delays, neither in the existing translation or in the English source strings.xml file, which means that there are two entries in Etar's settings ("Default snooze delay time" and "Ask for reminder snooze delay") that I am currently unable to translate.

DandelionSprout:
Nevermind, I found them. Apparently they were hiding around in cmstrings.xml.

res/values-nb/strings.xml
```diff
@@ -274,4 +284,17 @@
     <item quantity="other" msgid="8779598489398048259">"For <xliff:g id="EVENT_COUNT">%d</xliff:g> aktiviteter"</item>
   </plurals>
     <string name="acessibility_recurrence_choose_end_date_description" msgid="5899029840837836711">"endre sluttdato"</string>
+    <item quantity="one">Hver dag</item>
+        <!-- Plural form [CHAR LIMIT=30] -->
+        <item quantity="other">Hver <xliff:g id="count">%d</xliff:g>. dag</item>
+    </plurals>
+    <plurals name="Ndays">
```
Gitsaibot: Twice, please remove this and do your changes in line 44...

res/values-nb/strings.xml
```diff
@@ -274,4 +284,17 @@
     <item quantity="other" msgid="8779598489398048259">"For <xliff:g id="EVENT_COUNT">%d</xliff:g> aktiviteter"</item>
   </plurals>
     <string name="acessibility_recurrence_choose_end_date_description" msgid="5899029840837836711">"endre sluttdato"</string>
```
Gitsaibot: Missing <plurals ....>

res/values-nb/strings.xml
```diff
@@ -284,10 +284,11 @@
...
+  <plurals name="recurrence_interval_daily">
```
Gitsaibot: This plural tag already exists in line 261. There you can make your changes ;)

DandelionSprout:
Ah, thanks for the headsup. My knowledge of coding is so-so, and as such I can't easily spot coding error, so thanks for walking me through this. 😊

DandelionSprout:
Indeed I've forgot what I've changed in the Ndays plural, and therefore I also can't remember how that string is relevant to any of the other strings I've edited... 😓

DandelionSprout:
No, wait, now I realise what you were hinting to. If I remember correctly, I figured out that the edits I did to Ndays in line 291 before I removed those duplicate strings, were not necessary to do after all. 🙂

PR #321, opened 2018-05-13
Add an intermediate dark theme, rename the old dark theme to "black"

TimoWilken:
Many other Android apps have a dark grey *dark* theme and a completely black AMOLED-friendly *black* theme. This duplicates the existing "dark" theme, adding a more standard dark grey theme.

This has the same shortcomings (some unthemed elements remaining white) as the existing "dark" theme, but I hope that it can be built upon.

# Screenshots

| Existing dark / new black theme | New dark theme |
|:-:|:-:|
| old dark theme / new black theme | new dark theme |


xsoh:
Thanks @TimoWilken for this PR!
I think it is better to call it `Gray`  so user doesn't get confused between names

TimoWilken:
Hi @xsoh, thanks! I don't particularly mind what it's called, though there seems to be a convention of using "light"/"dark"/"black" in a particular way. For instance, here is Orgzly:

| "Light" theme | "Dark" theme | "Black" theme |
|:-:|:-:|:-:|
| screenshot_20180525-161745 | screenshot_20180525-161751 | screenshot_20180525-161756 |

I'm happy to call it "gray" if you'd like me to. I admit that it might be confusing for users if the name is used by a different theme.

jspricke:
+1 for the naming scheme proposed in the patch.

jspricke:
@Gitsaibot @xsoh can we merge this?

TimoWilken:
I've now used `DynamicTheme` wherever `Utils.getTheme` was used before (@Gitsaibot).

TimoWilken:
I personally prefer the lower contrast. It seems comparable to the contrast of the light-grey-on-white lines in the light theme. I think the problem is that the past is highlighted in a fairly light grey compared to the normal background. Would making the highlighting a bit darker be acceptable? I'm happy to make the grid lines a bit lighter as well, I wouldn't want to overdo it though.

xsoh:
@jspricke  I've tested it 👍   I'm good with merging it 👍 .. Thanks @TimoWilken !

jspricke:
@Gitsaibot please remove the master merge ;).

res/values-fr/arrays.xml
```diff
@@ -167,11 +167,6 @@
     <item msgid="5499975799312474425">"tous les derniers samedis"</item>
   </string-array>
 
-    <string-array name="pref_theme_entries">
```
jspricke: Why should this be removed?

Gitsaibot: It's not necessary. We do the translation in @string/preferences_black_theme...

jspricke: I see, fine with me.

PR #361, opened 2018-08-28
Don't save event if onPause() is triggert

### LINKED ISSUE #216: New items created twice, should update item after app switch
Etar creates items twice in this situation:
- start creating a new item, fill in some details
- go to another app (e.g. your browser to copy event description or link)
- go back to Etar to paste the stuff, tap save

In that case the item is stored twice: once without and once with the copy-and-pasted info. Instead, I would expect Etar to remember it still had the thing open and that it should update the item rather than creating a new one.

Etar 1.0.10 (F-Droid)
Device: Fairphone 2
OS: Fairphone OS 17.04.8 (Android 6.0.1)

Gitsaibot:
The use of the homebutton during a event creation shouldn't trigger a save event. Fixes #216

### DIFF
--- a/src/com/android/calendar/event/EditEventFragment.java
+++ b/src/com/android/calendar/event/EditEventFragment.java
@@ -568,12 +568,6 @@ public class EditEventFragment extends Fragment implements EventHandler, OnColor
 
     @Override
     public void onPause() {
-        Activity act = getActivity();
-        if (mSaveOnDetach && act != null && !mIsReadOnly && !act.isChangingConfigurations()
-                && mView.prepareForSave()) {
-            mOnDone.setDoneCode(Utils.DONE_SAVE);
-            mOnDone.run();
-        }
         super.onPause();
     }
 


jspricke:
>Yes you are right. But at the moment I don't know how to resolve this. Besides who open Etar with the icon if it is already open ?

I think both methods should result in the same behaviour. At least for 
me the Etar button on the desktop is often easier to find then scrolling 
through the apps.

>In my opinion it is better not to auto save an event like Etar does. No user would expect a save event(without asking) when he press the home or back button.

Well, I think saving content is better then deleting it. For example K-9 
mail saves an edited draft message if you pause it.

An other option would be to make the content more persistent (I'm not 
so fluent in the Android world, but some quick Google turned up this: 
https://stackoverflow.com/questions/151777/saving-android-activity-state-using-save-instance-state)
So we store the user data, recover it and only save it to the calendar 
once the user presses the button.

Also, I think with this patch the content would be lost if android would 
kill the Etar process in the background, which could happen quiet fast 
on devices with low memory.

As a quick solution, we could probably just close the edit view upon 
pause, so the event gets saved and the user has to open it again upon 
return.

PR #365, opened 2018-09-18
Untoggle "All day" in Monthview leads now to a 1h long event 

### LINKED ISSUE #289: Untoggle "All day" leads to a 24h long event
Steps to reproduce:

1. Long press a day
2. Press Edit
3. Untoggle "All day"

Expected:
Event starting on the day I pressed and ends on the same day

Actuall:
Event starting on the day I pressed at 0:00 and ends next day at 0:00

Gitsaibot:
Previously untoggle an "All day" event in Monthview led to a 12 hour event. Now it is only a one hour long event.  Fixes #289

### DIFF
--- a/src/com/android/calendar/event/CreateEventDialogFragment.java
+++ b/src/com/android/calendar/event/CreateEventDialogFragment.java
@@ -144,7 +144,7 @@ public class CreateEventDialogFragment extends DialogFragment implements TextWat
                         public void onClick(DialogInterface dialog, int which) {
                             mController.sendEventRelatedEventWithExtraWithTitleWithCalendarId(this,
                                     EventType.CREATE_EVENT, -1, mDateInMillis,
-                                    mDateInMillis + DateUtils.DAY_IN_MILLIS, 0, 0,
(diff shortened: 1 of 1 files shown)


jspricke:
For me it leads to a 1h long event (not a 12h as the subject is saying). I guess that's the intention as well.
 I would have expected a +3600000 here, is this -1 for one hour somewhere documented or are we using some flaw in the API here?

jspricke:
As you are editing the commit anyhow, some comments to it:

The original title was fine, just make it 1h instead of 12h ;). The new one needs much more context to understand it.
Put more information in the body of the commit comment. The format is to have the first line as a Title of the commit, then an empty line and then more text. So for example the title could be "Untoggle "All day" in Monthview leads now to a 1h long event" (btw. some projects enforce a maximal length of the title, I would see it more like a soft constraint) and then have the rest of it ("..and not to a 24h event. Fixes #289"; as a full sentence of cause) in the body below the new line.

Also, my question regarding the -1 convention is still open, do you have a reference where it is done that way in the week view?

Issue #372, opened 2018-09-28
Notification doesn't work with target SDK 27+


xsoh:
Notification doesn't work  anymore with `targetSdkVersion 27`.

PR #375, opened 2018-09-28
 Fix: Notification doesn't work with target SDK 27+ #372 

xsoh:


src/com/android/calendar/alerts/AlertReceiver.java
```diff
@@ -319,7 +327,7 @@ private static Notification buildBasicNotification(Notification.Builder notifica
                         resources.getString(R.string.snooze_label), snoozeIntent);
                 numActions++;
             }
-            return notificationBuilder.getNotification();
+            return notificationBuilder.build();
```
jspricke: Is this backward compatible?

jspricke:
@Gitsaibot you can edit the notification in the app properties and add a LED light there as well.

jspricke:
Try setting the importance to urgent (Wichtigkeit: dringend).

PR #413, opened 2018-12-29
Fix 404 - move notification channel creation to AllInOneActivity.

xsoh:


Catfriend1:
Thanks for fixing the issue, will try it on next release.

PR #415, opened 2019-01-04
Prevent white screen at startup.

### LINKED ISSUE #409: Quick white screen at startup
Hello,
I use the dark theme and for me this app is perfect apart from maybe just the white screen that flashes the eyes at the start of the application.

But otherwise it works very well and I find the visual and the ergonomics intelligently thought.
Congratulations and thank you very much !

Gitsaibot:
Fix #409


jspricke:
This feels like I missed the icon upon start because the screen is not changing. I would prefer a black background as implemented in #416 .

PR #429, opened 2019-01-21
Event formatting in month view refactored to introduce EventFormatter

0xym:
Long term goal of my changes is to allow multi-line calendar events in MonthView (this should be configurable)
This commit decouples dealing with some formatting issues - like whether to show event time, or "more" indicator, from actual event drawing.

### DIFF
--- a/src/com/android/calendar/month/MonthWeekEventsView.java
+++ b/src/com/android/calendar/month/MonthWeekEventsView.java
@@ -1510,8 +1510,19 @@ public class MonthWeekEventsView extends SimpleWeekView {
         NullFormattedEvent(EventFormat format, BoundariesSetter boundaries) {
             super(format, boundaries);
         }
+
+        /**
+         * Null object has no text to be formatted
+         */
         public void initialPreFormatText() {}
         protected boolean areTimesVisible(boolean showTimes) { return false; }
+
+        /**
+         * Null object won't be drawn
+         * @param canvas
+         * @param showTimes
+         * @param day
+         */
         public void draw(Canvas canvas, boolean showTimes, int day) {}
     }
 
--- a/src/com/android/calendar/month/MonthWeekEventsView.java
+++ b/src/com/android/calendar/month/MonthWeekEventsView.java
@@ -35,7 +35,6 @@ import android.graphics.Paint.Style;
 import android.graphics.Typeface;
 import android.graphics.drawable.Drawable;
 import android.provider.CalendarContract.Attendees;
-import android.support.annotation.Nullable;
 import android.text.DynamicLayout;
 import android.text.Layout;
 import android.text.TextPaint;
@@ -756,16 +755,15 @@ public class MonthWeekEventsView extends SimpleWeekView {
 
     protected class DayEventSorter {
         private final EventFormat virtualFormat = new EventFormat(0,0);
-        private LinkedList<FormattedEvent> mRemainingEvents;
+        private LinkedList<FormattedEventBase> mRemainingEvents;
         private BoundariesSetter mFixedHeightBoundaries;
-        private FormattedEvent mVirtualEvent;
+        private FormattedEventBase mVirtualEvent;
         private int mListSize;
         private int mMinItems;
         public DayEventSorter(BoundariesSetter boundariesSetter) {
             mRemainingEvents = new LinkedList<>();
             mFixedHeightBoundaries = boundariesSetter;
-            mVirtualEvent = new FormattedEvent(null, virtualFormat);
-            mVirtualEvent.setBoundaries(boundariesSetter);
+            mVirtualEvent = new NullFormattedEvent(virtualFormat, boundariesSetter);
         }
 
         /**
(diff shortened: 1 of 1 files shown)


0xym:
I suppose it should be ready for testing now.
Do you have some automatic test suite? I haven't run any unit tests (nor found any). I have tested some obvious scenarios and it seems to me that everything works the same way. 
I also don't know what is your merging strategy - do you want to merge my chances only after there will be some functional changes?

jspricke:
Thx for your pull request. You are changing quiet a bit of code, so I will probably need some more time to review it. I played around with it a little andfound one bug, can you fix that maybe?
etar

0xym:
Interesting. Unfortunately I won't have time to look into it until Tuesday. I hope it will be easy to reproduce it.

0xym:
I have started the work on multiline events on Friday. I've had initial code ready yesterday. Today I have fixed remaining bugs. It turns out that the bug you have mentioned should be fixed with this commit.
To turn off multiline events it is enough to set mMaxLinesInEvent to 1 (I have set it to 7 to so it would be almost impossible that it would be a real limit - but I didn't want to set it to a higher value, because increasing it doesn't come for free).

0xym:
the issues should be gone now

jspricke:
I found an other instance of showing less events then available space. I have two full day events with a short title and three non full day. The first non full day has a rather long title (>47 chars in my case). The is that the two full day events and a `+3` are shown. If I cut the title at 46 chars, the beginning of it is shown as well (and a '+2'). I didn't went through the code yet, but maybe you have an idea why this happens?

Really nice work otherwise!

0xym:
I have looked briefly into the code and I think that t there is "off by one" error in line 1176. In reduceHeightOfEventsToOne the loop should be inclusive (condition should be "<=" )

jspricke:
@0xym the "<=" solved it, thx! I pushed a fixup.

0xym:
I suppose that this part of changes is complete now (unless there are some bugs). There is one thing that might need to be added - some configuration. But I don't know what should be configurable, to what extent and where to put the configuration.

0xym:
For dealing with text formatting I'm using android.text.DynamicLayout. Text breaking can be improved, but only in API28 (Android 9) - there is DynamicLayout.Builder which can produce DynamicLayouts with selected breaking strategies (there are 3 to choose from now). Currently less than 1% of Android devices runs Pie, so I'm afraid that going this way won't make a big difference

src/com/android/calendar/month/MonthWeekEventsView.java
```diff
@@ -103,9 +106,12 @@
     private static int EVENT_BOTTOM_PADDING = 1;
     private static int TODAY_HIGHLIGHT_WIDTH = 2;
     private static int SPACING_WEEK_NUMBER = 0;
+    private static int BORDER_SPACE;
```
jspricke: I think these should be lower case (as they are not constants), see: https://source.android.com/setup/contribute/code-style#follow-field-naming-conventions

(also it's not consistent in the file).

src/com/android/calendar/month/MonthWeekEventsView.java
```diff
@@ -785,152 +753,912 @@ protected int addChipOutline(FloatRef lines, int count, int x, int y) {
         return count;
     }
 
+    protected class DayEventSorter {
+        private final EventFormat virtualFormat = new EventFormat(0,0);
```
jspricke: add a space after the comma.

src/com/android/calendar/month/MonthWeekEventsView.java
```diff
@@ -785,152 +753,912 @@ protected int addChipOutline(FloatRef lines, int count, int x, int y) {
...
+                        break;
+                    }
+                    index = getNextIndex(indexedEvents, index);
+                }
+            }
+            addVirtualEvents(indexedEvents, index);
+        }
+
+        protected void addVirtualEvents(FormattedEventBase[] indexedEvents, int initialIndex)  {
+            for (int index = initialIndex; index < mMinItems; ++index) {
```
jspricke: I would go for `index++` (seems to be more common in the code base).

src/com/android/calendar/month/MonthWeekEventsView.java
```diff
@@ -785,152 +753,912 @@ protected int addChipOutline(FloatRef lines, int count, int x, int y) {
...
-         *
-         * e. "+ more" line (mExtrasHeight)
-         *
-         * f. EVENT_BOTTOM_PADDING (overlaps EVENT_LINE_PADDING)
-         *   (bottom of box)
+    protected class DayEventFormatter {
+        private ArrayList<FormattedEventBase> mEventDay;
+        private boolean mShowTimes;
+        private int mDay;
+        //info
```
jspricke: What should the comment mean?

src/com/android/calendar/month/MonthWeekEventsView.java
0xym: fixed

src/com/android/calendar/month/MonthWeekEventsView.java
0xym: done

src/com/android/calendar/month/MonthWeekEventsView.java
0xym: using pre incrementation is more efficient (but compiler should figure it out and change post-incrementation to pre-incrementation whenever it has no side effect)
I have changed the code to follow your suggestion

src/com/android/calendar/month/MonthWeekEventsView.java
0xym: comment changed to a more meaningful one

PR #440, opened 2019-02-09
Remove unused file

Gitsaibot:


jspricke:
can you add the complete Apache license as well?

jspricke:
I would have said http://www.apache.org/licenses/LICENSE-2.0. My point was rather to have rm MODULE_LICENSE_APACHE2 and adding the real license in one commit.

PR #443, opened 2019-02-14
Month view configuration

0xym:
This code is based on my previous changes.
It is not ready yet, but shows the concept.

### DIFF
--- a/res/menu-land/all_in_one_title_bar.xml
+++ b/res/menu-land/all_in_one_title_bar.xml
@@ -52,4 +52,10 @@
         android:alphabeticShortcut="h"
         android:title="@string/hide_controls"
         app:showAsAction="never" />
+    <item
+        android:id="@+id/action_view_settings"
+        android:alphabeticShortcut="v"
+        android:title="View settings"
+        android:orderInCategory="8"
+        app:showAsAction="never" />
 </menu>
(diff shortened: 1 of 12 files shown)


0xym:
I have added initial implementation of time formatting. It seems that there is no good way to format duration at the moment, so I might hide this option for the moment.
I'm also wondering shouldn't there be more options with the time below (like to have just start time - and to allow it also on a small (non tablet) devices)? Maybe there should be separate control for tie below and inline?

0xym:
> The settings apply only to the month view right? Why not use the 3 dots in the right corner(in month view) to get to the view settings? Just an idea. Two clicks against 4 ;)

I'll try to do as you suggest.
I also think that it would be better to move "About Etar" one level up, so the settings would correspond to "General settings" (I can't see other settings than "General settings and "About Etar" in settings, so I think that it would be good). Maybe I'll do it on a separate branch.

0xym:
done

0xym:
I'm having some issues with rebase, but I hope to solve them soon (the issue is with building latest Etar-Group:master from scratch so it is rather some local issue)

0xym:
actually this is the error I'm getting now

0xym:
well `git submodule update --init` results in:
error: Server does not allow request for unadvertised object 12b1f71df3969c976dc520f3048efa7660532aad
Fetched in submodule path 'external/datetimepicker', but it did not contain 12b1f71df3969c976dc520f3048efa7660532aad. Direct fetching of that commit failed.

jspricke:
@0xym should be fixed, can you retry?
I had the rebase branch in datetimepicker already forwarded to the new master (using androidx). Reverted that for now.

0xym:
it works now.
I found it easier to squash the commits and cherry-picked them to master. If you are ok then I'll proceed with new pull request

PR #516, opened 2019-06-15
Remove app:theme and use android:theme instead.

Gitsaibot:
See log message: app:theme is now deprecated. Please move to using android:theme instead.


jspricke:
Are you sure the cleanup doesn't remove code from the deactivated widget (in case we want to revive it again at some point)?
Otherwise fine with me.

Issue #511, opened 2019-05-31
Feature Request: Be able to cut off words of events and change font size


realmain:
Being able to have all event names cut off in month view and reduce the font size of the day number & event name would help reduce the clutter look. Etar currently only does this if there are too many events in one day, would be nice if it can toggle it to **always** do it instead.

e.g.

**Etar**
<img src="https://i.imgur.com/cWR3Br4.png" width="50%" height="50%" />

**Simple Calendar**
<img src="https://i.imgur.com/Vu3kffT.png" width="50%" height="50%" />

Gitsaibot:
> Being able to have all event names cut off in month view; Etar currently only does this if there are too many events in one day, would be nice if it can toggle it to always do it instead.

@0xym Is that an easy thing to do? Personally, I think it's good the way it is but I have nothing against an option to make this configurable.

0xym:
https://github.com/Etar-Group/Etar-Calendar/blob/b248a37bbbdc74c7302937e5d8d1cbcfd7131a70/src/com/android/calendar/month/MonthWeekEventsView.java#L115

I guess that this line answers all questions :)

0xym:
BTW - number of lines should not be huge, because increasing it has impact on performance

Gitsaibot:
Thank's, that is the line. Now I just need time to get the job done ;) Or maybe someone else is faster...

0xym:
ok - I'm working on it now

Gitsaibot:
Solved. Thanks at @0xym

PR #528, opened 2019-08-02
Max number of lines in event made configurable

0xym:
I have fixed Polish translations to better match the context and intention of settings.
But I have also changed the text of "show_time_no" (I'm not sure if it was the right thing to do, but I suppose that it looks better now).

I also had to add a mechanism to allow updating of xml files containing default settings (previously if the file was updated, the default value wound not be loaded into shared prefs, simple changing the "readAgain" to true would on the other hand force it to parse the file on every re-setart).

### DIFF
--- a/res/values-pl/strings.xml
+++ b/res/values-pl/strings.xml
@@ -342,9 +342,10 @@
     <string name="two_hours">2 godziny</string>
     <string name="view_settings">Ustawienia widoku</string>
     <string name="default_start_title">Widok domyślny</string>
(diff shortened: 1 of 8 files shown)


0xym:
#511

0xym:
done

0xym:
I have chosen 7 as an arbitrary value, but as a matter of fact I can fit 7 line event on my screen.
I guess that I need to check how many lines can fit to cap available values (I was hoping I could skip it for first version)

0xym:
done

Issue #431, opened 2019-01-26
No Labels visible in multi-day appointments


spaetz:
In many multiday appointments Etat does not put Labels in the appointment although there would be enough space. For instance multi-week appointments should have their labels repeated.on every week when looking at the in the weekly view. I will try to attach an screenshot demonstrating the issue.

spaetz:
Sorry, cannot upload img directly from Android to github. Screenshot is here:
https://pasteboard.co/HYdqIjx.png

Gitsaibot:
Can't reproduce this with current master

spaetz:
All I can say is that the latest available version on f-droid (1.0.14) had/has the issue. I would be very much willing to recheck on a newer release.

Gitsaibot:
Does this still happen with the current release?

spaetz:
![Screenshot_20190709-075648](https://user-images.githubusercontent.com/63132/60862981-8de82d00-a21f-11e9-8eb1-4e0bc7af5220.png)
This is etar 1.0.17. the all-week labels are "Schulferien" and "Sommerurlaub" (so they should not be too long to be displayed on the screen). So yes, it is still happening :-)

Gitsaibot:
Does this also happen in month view? Have you created the multi-day appointments with etar or have they been imported?

spaetz:
No, it does not happen in month view, month view looks fine.

I have events that are created externally (Exchange), but I just tested with a Week long event created with Etar and it happens with that one too.

Gitsaibot:
That's weird. Right now I have no idea why this is happening. Maybe @jspricke or @0xym have an idea...

0xym:
I was able to reproduce it when I have set font size (in android settings) to maximum value.
It seems that the reason is that minimum event height is not set to the right value. It is visible when there are more than 3 all day events displayed in a single day.

spaetz:
Ohh, yes, we are on to something here.
My font size was set to large, not standard (it was not set to the biggest possible size).

Reducing font size to normal makes all labels appear. It also only happens when there are 3 or more all day appointments, how could I have failed to realize that!

jspricke:
fixed in master

PR #544, opened 2019-08-22
Fixed issue with wrong height of events in week view

0xym:
I'm not sure what about event_min_height dimension - should we keep it, or should we use a real minimum height instead.

### DIFF
--- a/res/values-w600dp-h879dp/dimens.xml
+++ b/res/values-w600dp-h879dp/dimens.xml
@@ -26,7 +26,6 @@
     <dimen name="week_view_event_text_size">12sp</dimen>
     <dimen name="event_text_vertical_margin">4dp</dimen>
     <dimen name="event_text_horizontal_margin">6dp</dimen>
-    <dimen name="event_min_height">30dip</dimen>
     <!-- minimum size of hours column in day/week view -->
     <dimen name="min_hours_width">32dip</dimen>
     <!-- margins size of hours column in day/week view -->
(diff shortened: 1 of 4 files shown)


0xym:
Fixes #431

0xym:
As far as I know "dip" is not scalable, so using it won't solve the issue in the long run. Also padding is not scaling with the font size, so "sp" won't produce the right values for all cases.

jspricke:
Some events are shown too small now even though there is enough space. Can you have a look?
Screenshot_20190906-104247_Etar
Screenshot_20190906-104313_Etar

jspricke:
It seems to only happen upon first start (after installing Etar). Restarting make the event big enough to read the description. The event is only 10 minutes long, could be related.

jspricke:
Played around a little:
* It's broken upon every app start (not only after installing Etar).
* Setting MIN_EVENT_HEIGHT the old way but in init() results in the same broken behavior. -> So we need to set it early.
* Computing fontHeight for the first time gives a really small value (17.0 compared to 66.0 for MIN_EVENT_HEIGHT) -> I guess we have a race condition.

This worked for me, not sure if it makes sense:
```
        float event_min_height = mResources.getDimension(R.dimen.event_min_height);
        if(fontHeight < event_min_height) {
            fontHeight = event_min_height;
        }
```

0xym:
I have looked into it and the were two issues:
mAnimateDayEventHeight was set to MIN_UNEXPANDED_ALLDAY_EVENT_HEIGHT, before MIN_UNEXPANDED_ALLDAY_EVENT_HEIGHT was initialized to the right value (but error didn't have any immediate adverse effect).
The culprit here was 
`        mEventGeometry.setMinEventHeight(MIN_EVENT_HEIGHT);
`
This line was called before the right value of MIN_EVENT_HEIGHT was set and it resulted in the observed issue.
I've found out that font is not used prior to init so I have moved it to the constructor - after all required values are computed.

jspricke:
Thanks, that looks much better :). I pushed away some unrelated whitespace changes and will test it the next days and merge then.

PR #548, opened 2019-08-29
Update from Weblate

weblate:


jspricke:
I'm not sure I get you here. If the commit on master is wrong, do a `git revert HEAD` and send a pull request. But it looks valid to me, so why can't Weblate adopt it?

PR #552, opened 2019-09-06
Sdk29

jspricke:


AndroidManifest.xml
```diff
@@ -37,7 +37,7 @@
     <uses-permission android:name="android.permission.WRITE_CALENDAR" />
     <uses-permission android:name="android.permission.WAKE_LOCK" />
     <uses-permission android:name="android.permission.READ_SYNC_SETTINGS" />
-    
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
```
jspricke: why do we need this?

Gitsaibot: Since api =28 **startForeground** requires **android.permission.FOREGROUND_SERVICE**
We use startForeground here: "Line 136 AlertReceiver.java"

PR #558, opened 2019-09-21
delete unused files in res

jspricke:


jspricke:
I guess it uses one of these:
external/chips/res/drawable-hdpi/ic_contact_picture.png
external/chips/res/drawable-mdpi/ic_contact_picture.png
external/chips/res/drawable-xhdpi/ic_contact_picture.png
external/chips/res/drawable-xxhdpi/ic_contact_picture.png
(but not sure, at least it compiles without)
I can add it again if you prefer.

Issue #520, opened 2019-07-05
Foreground service notification


ByteHamster:
The app recently got a foreground notification for the AlarmService. This is quite distracting because I now receive 2 heads up cards per event. One that actually is about the event and one that just tells me that the service is running. I suggest to change the behavior so that only one notification is displayed. For this purpose, I have 2 possible solutions.

- Post the foreground notification on a different channel, so I can hide it in the Android settings (quick and dirty).
- Convert the AlarmService to a JobIntentService. It looks like the service is only started for a very short period of time. As soon as its work is done, the service is stopped. This is exactly the use-case of a Job instead of a Service. The JobIntentService class from the support libraries automatically uses a Job on new Android versions and therefore does not need a foreground notification. This way would be more clean. https://medium.com/til-kotlin/jobintentservice-for-background-processing-on-android-o-39535460e060

jspricke:
Thanks for the analysis, can you come up with a pull request? Given the age of the code, I can imagine that your second option is correct (didn't look into the details), but trying your first option first would be fine with me as well.

ByteHamster:
> can you come up with a pull request?

Hmm I am quite busy currently. Let's see when this is bugging me enough to clone the repo :P

Gitsaibot:
> Convert the AlarmService to a JobIntentService

That's the way we should go. Besides this would solve #521 which is caused by using the same channelID, I think.

Cwpute:
I am still having this problem. Has the issue been fixed for you since then @ByteHamster ?

PR #559, opened 2019-09-22
move foreground notification to own channel

michaelPotter:
Since the android 10 update rolled out, my phone buzzes every time the foreground service runs and it's been driving me bonkers. This commit moves the foreground service notification to its own notification channel and sets the priority to low, so it shouldn't make noise anymore. You can also hide the notification completely this way.
I did look into the JobIntentService solution from #520, but wasn't able to get that working.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -248,6 +248,10 @@
     <string name="alert_title">Calendar notifications</string>
     <!-- Notification window messages: -->
     <skip/>
+    <string name="foreground_notification_channel_name">Calendar Sync</string>
+    <string name="foreground_notification_channel_description">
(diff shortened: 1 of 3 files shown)


michaelPotter:
Huh. It works on my phone (pixel 2, android 10) and an emulator running android 9. I added a third channel to test that they will get added right, and that worked too.
Are you on the right branch? It's not on my fork's master.
Also, what kind of phone do you use?

michaelPotter:
Hey @Gitsaibot  I've renamed the channel and description to your preferred wording

dtzWill:
Oh thank you, this was driving me insane! Please consider merging for next release, etar powers that be! :)

PR #574, opened 2019-10-22
create notification channels before using

### LINKED ISSUE #517: android.app.RemoteServiceException
Since last release we get this error on android play console:

> android.app.RemoteServiceException: 
>   at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1821)
>   at android.os.Handler.dispatchMessage (Handler.java:106)
>   at android.os.Looper.loop (Looper.java:164)
>   at android.app.ActivityThread.main (ActivityThread.java:6626)
>   at java.lang.reflect.Method.invoke (Native Method)
>   at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:438)
>   at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:811)

This only affects android 8.1 and 9, it seems. Need a full debug log...

michaelPotter:
Moves notification channel creation into AlertService.java so it can be done before sending foreground notifications.
Fixes #517

### DIFF
--- a/src/com/android/calendar/alerts/AlertService.java
+++ b/src/com/android/calendar/alerts/AlertService.java
@@ -984,6 +984,8 @@ public class AlertService extends Service {
         if (intent != null) {
 
             if (Utils.isOreoOrLater()) {
+
+                createChannels(this);
                 Notification notification = new NotificationCompat.Builder(this, FOREGROUND_CHANNEL_ID)
                         .setContentTitle("Event notifications")
                         .setSmallIcon(R.drawable.stat_notify_calendar)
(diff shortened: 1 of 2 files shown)


michaelPotter:
The channels will still be created in AllinOneActivity (w/call to AlertService.createChannels() ) I just moved the code into a function so we're not having multiple copies all over

src/com/android/calendar/alerts/AlertService.java
```diff
@@ -1010,6 +1012,29 @@ public IBinder onBind(Intent intent) {
         return null;
     }
 
+    public static void createChannels(Context context) {
+        if (Utils.isOreoOrLater()) {
+            // Create notification channel
+            NotificationMgr nm = new NotificationMgrWrapper(
+                    (NotificationManager) context.getSystemService(Context.NOTIFICATION_SERVICE));
+
```
Gitsaibot: Maybe we should check if the channels already exists ?

michaelPotter: Oh wow. I'm sorry. Apparently this comment was "pending" all week...

I don't think there is a need to check. The api seems to be designed to no-op if the channel already exists
>Creating an existing notification channel with its original values performs no operation, so it's safe to call this code when starting an app.

https://developer.android.com/training/notify-user/channels#CreateChannel

PR #576, opened 2019-10-24
Avoid VEvent to erase key if exists

theo-dep:
Correct a problem with *.ics shared by ICal. The file has many DTSTART key (see attachment) but only the first is correct.

[iCal-2019930.ics.txt](https://github.com/Etar-Group/Etar-Calendar/files/3767547/iCal-2019930.ics.txt)


jspricke:
Hi, thanks for your pull request. I looked into your .ics file and while there are multiple lines with DTSTART, they each belong to a differnt element (i.e. different VEVENT, VTIMEZONE.... So technically the file is correct, as there can be only one DTSTART per element, according to the RFC. So the problem is rather why Etar didn't stop at the end of the VEVENT. The real fix is in #585.

theo-dep:
Thanks 🙂

PR #587, opened 2019-11-01
weblate translation error corrected

Gitsaibot:


jspricke:
Doing
```git grep "\\\n"```
finds some more of these. Do you plan to work on them or should I work on them?

jspricke:
@Gitsaibot this should be all, can you have a quick look?

PR #605, opened 2019-12-07
Add gradle wrapper

d7415:
Add a gradle wrapper, so that a different gradle version can be used for Etar than the system-wide version.

This might not be perfect (I don't really know gradle) but it works.


jspricke:
I think it's not the job of the Etar repository to work around old gradle versions in the OS and I would rather like to see Etar supporting old gradle versions (a stable gradle would be even better). This adds a binary .jar which I don't know the license nor origin of and can't verify. Also it adds a specific version in the properties file, which needs to be maintained. We agreed in #106 that we don't need it, but if there are new circumstances such that we should rethink the decision, I would be open.
Also, if we decide to add the wrapper, what about the gradlew.bat?

d7415:
> I think it's not the job of the Etar repository to work around old gradle versions in the OS and I would rather like to see Etar supporting old gradle versions (a stable gradle would be even better). This adds a binary .jar which I don't know the license nor origin of and can't verify. Also it adds a specific version in the properties file, which needs to be maintained. We agreed in #106 that we don't need it, but if there are new circumstances such that we should rethink the decision, I would be open.
> Also, if we decide to add the wrapper, what about the gradlew.bat?

Good points.

* I didn't have the correct gradle version to hand, so had to steal and adapt one from elsewhere. Chicken and egg problem that I didn't have a supported version, and because of that was unable to generate the wrapper. Generating it specifically using a supported version of gradle would be a better option, and would include gradlew.bat and anything else I missed.

* I hadn't seen #106, but it does seem to be the norm to include the wrapper with apps.

* The version in properties was just the version it told me I needed when I used my system version (4.4.1).

That seems enough for me to close this. I'll use this locally until/unless a wrapper is added on the project side.

PR #613, opened 2020-01-03
Remove external datetimepicker

### LINKED ISSUE #600: Date picker size
This is with a small screen device - date picker renders over the viewport, so I cant pick sundays, eg. when editing an event

Screenshot_20191123-233840 Screenshot_20191123-233859

Reducing Display / Font Size does not help, adb shell density does, but I rather not use it

### LINKED ISSUE #603: When in dark mode: selecting a time via the clock graphic is still bright
The analog clock is really bright. Is this possible to dim when in a dark theme?

Gitsaibot:
Fix #603 ; Maybe Fix #600


hungerburg:
@Gitsaibot Hello there; I see you squash two flies at once - Just a user here, before going on a compile spree, though: I consider using system provided widgets usually wise.

Looking up on the android.app.datepickerdialog class - one can do themes with it - so that might help with dark mode?

Probably, for later, would be nice to have a preference for small screen devices, where the /spinner-mode/ might be a better choice.

Here someone sampled the class - https://www.zoftino.com/android-datepicker-example

Peter

hungerburg:
Well then, I checked out your fork in studio and had a build up on the emulator in, say 10 minutes? The datepicker looks just the same though? Fine, I am on the wrong branch. Not so easy. The other one is not a local branch acc. to studio. So I try and rebase and after some cleaning: Rebase Successful: Fast-forwarded master to origin/datetimepicker

Tada! I get a spinner!

Screenshot_1578170866

res/color/recurrence_bubble_text_color.xml
```diff
@@ -23,4 +23,4 @@
           android:color="@color/done_text_color_disabled"/>
     <item
           android:color="@color/recurrence_bubble_text_normal"/>
-</selector>
+</selector>
```
jspricke: nit: no need to change this ;).

src/com/android/calendar/event/EditEventView.java
```diff
@@ -1726,44 +1701,36 @@ public void onClick(View v) {
 
             TimePickerDialog dialog;
             if (v == mStartTimeButton) {
-                mTimeSelectedWasStartTime = true;
```
jspricke: same here and below.

src/com/android/calendar/event/EditEventView.java
```diff
@@ -377,28 +373,7 @@ public void onNothingSelected(AdapterView<?> arg0) {
         if (tzpd != null) {
             tzpd.setOnTimeZoneSetListener(this);
         }
-        TimePickerDialog tpd = (TimePickerDialog) fm.findFragmentByTag(FRAG_TAG_TIME_PICKER);
-        if (tpd != null) {
-            View v;
-            mTimeSelectedWasStartTime = timeSelectedWasStartTime;
```
jspricke: Why are you dropping this? (Haven't looked what it is used for exactly but timeSelectedWasStartTime Is a constructor parameter.)

Gitsaibot: This is stuff from the old datetimepicker. I don't think we'll be needing this anymore.

res/color/recurrence_bubble_text_color.xml
Gitsaibot: True ;)

hungerburg:
@Gitsaibot Never mind. This is perfectly fine, as seen in the screenshot; PS: The emulator runs API 29.

dschuermann:
Great work! I think it makes a lot of sense to get rid of these external dependencies and switch to platform dialogs or official support libraries where possible.

jspricke:
I would propose to move the "About Etar" to the toolbar and add a divider after "General settings" (maybe even with a "Accounts" header).

PR #615, opened 2020-01-06
Change License to GPLv3

Gitsaibot:


jspricke:
We may want to keep the old license around as `LICENSE.apache2`?

P.S.: Please push branches to your fork next time ;).

jspricke:
* Gitsaibot <notifications@github.com> [2020-01-05 23:59]:
>> We may want to keep the old license around as LICENSE.apache2?
>
>Isn't it a little confusing

Part's of the project stay with Apache 2, so we should provide the 
license file for that. Adding GPL 3 will make the combination GPL 3 but 
if someone copies just Apache 2 licensed code, it will stay Apache 2 
(just as we got it from the original authors, basically).

>> P.S.: Please push branches to your fork next time ;).
>
>I did this direct with my phone browser via Github ;) Didn't find a way to push it to my fork.

Not sure how that works but I would like to keep the main repo clean of 
branches because whoever pulls form it will get them and they are not 
cleaned up automatically. Having a fork adds a private namespace.

jspricke:
I cleaned this up a little:
- squashed to one commit.
- don't modify LICENSE.apache2 while coping it.
- use up to date version of GPL3 (with https address ;) ).

xsoh:
> Part's of the project stay with Apache 2, so we should provide the
> license file for that. Adding GPL 3 will make the combination GPL 3 but
> if someone copies just Apache 2 licensed code, it will stay Apache 2
> (just as we got it from the original authors, basically).
Shouldn't we mention the parts that under Apache 2 license?

jspricke:
>> Part's of the project stay with Apache 2, so we should provide the
>> license file for that. Adding GPL 3 will make the combination GPL 3 but
>> if someone copies just Apache 2 licensed code, it will stay Apache 2
>> (just as we got it from the original authors, basically).
>Shouldn't we mention the parts that under Apache 2 license?

Maybe change the `Readme.md` to:

Licensed under the GPLv3: https://www.gnu.org/licenses/gpl-3.0.html 
Except where otherwise noted.

PR #623, opened 2020-01-14
Redesign about dialog

Gitsaibot:


res/values/themes.xml
```diff
@@ -37,6 +37,9 @@
         <item name="calendar_future_bg_color">@color/calendar_future_bg_color_dark</item>
         <item name="month_other_bgcolor">@color/month_other_bgcolor_dark</item>
         <item name="settings_calendar_offline">@drawable/ic_sync_off_dark</item>
+        <item name="iconAboutAuthors">@drawable/ic_create_light</item>
```
dschuermann: The naming of drawables is currently the other way round ;) Light drawables are postfixed with "dark" as they are used in the dark themes.

src/com/android/calendar/settings/AboutPreferences.kt
```diff
@@ -17,32 +17,61 @@
...
 
-class AboutPreferences : PreferenceFragmentCompat() {
+import java.util.Calendar
+import ws.xsoh.etar.R
+import kotlinx.android.synthetic.main.about_preferences.*
 
-    override fun onCreatePreferences(savedInstanceState: Bundle?, rootKey: String?) {
-        setPreferencesFromResource(R.xml.about_preferences, rootKey)
-        setVersion()
+class AboutPreferences : Fragment() {
```
dschuermann: Because it's no longer a ``PreferenceFragmentCompat``, I would propose…
* moving it out of the settings package
* no longer calling it using the ``SettingsActivity``. Either put it in a new Activity or replace it inside the ``AllInOneActivity``

dschuermann:
We can also remove ``about_preferences.xml``

res/values/themes.xml
Gitsaibot: Correct. Was a little late yesterday ;)

src/com/android/calendar/settings/AboutPreferences.kt
Gitsaibot: Tried to put the `AboutFragment` right in `AllInOneActivity` . Unfortunately without success. So I created a new activity. If you know how to bind the fragment directly in, I'd like to adapt it.

dschuermann: Yeah, I think having it in a separate activity is fine for now. The ``AllInOneActivity`` should be refactored at some point following a modern architecture.

src/com/android/calendar/AllInOneActivity.java
```diff
@@ -1390,39 +1388,6 @@ public boolean onQueryTextSubmit(String query) {
         return true;
     }
 
-    //@Override
```
jspricke: Can you comment why this can be removed?

Gitsaibot: It is not used anywhere. Or am I wrong ?

jspricke: it overrides the method in `NavigationView` I assume, but I don't so any difference in the behavior of the app. @xsoh @dschuermann has one of you more insights here?

Gitsaibot: But @Override is commented out. Besides at line 927 there is the same call.

jspricke: I missed the other implementation:
> One of the most notable change here is the method:
> `onNavigationItemSelected(MenuItem item)`
> The old Navigation Drawer template's definition of this method was:
> `onNavigationItemSelected(int position, long itemId)`
https://stackoverflow.com/q/32944798
Looks like it's fine.

PR #624, opened 2020-01-20
Local calendars: Remove events from database, not just mark them deleted

dschuermann:
If an event is part of a local calendar, really remove it from the database by making the operation as a sync adapter.

Related issue: https://github.com/SufficientlySecure/calendar-import-export/issues/44

> There are two versions of delete: as an app and as a sync adapter.
> An app delete will set the deleted column on an event and remove all instances of that event.
> A sync adapter delete will remove the event from the database and all associated data."

from https://developer.android.com/reference/android/provider/CalendarContract.Events


dschuermann:
I was working on the issue of editing recurring events and then went in the direction of fixing the deletion of events 😂.

So #662 will be a different one.

PR #629, opened 2020-01-21
Add more icons

Gitsaibot:


src/com/android/calendar/settings/MainListPreferences.kt
```diff
@@ -242,6 +247,16 @@ class MainListPreferences : PreferenceFragmentCompat() {
...
+    override fun onResume() {
+        super.onResume()
+        val context = preferenceManager.context
+        val screen = preferenceManager.createPreferenceScreen(context)
+
+        addGeneralPreferences(screen)
+
+        preferenceScreen = screen
+    }
+
```
Gitsaibot, line 259: Not sure if this is the best way to update the fragment state. It seems that the screen now flickers when switching between the fragments @dschuermann Maybe you know a better way to update the icons ;)

jspricke: Why do we need this?

Gitsaibot: To solve the glitch I mentioned above:

> If you switch from light theme to dark/black theme, the icon color (for general preferences) has not changed. You have to leave the settings. The other way from black to light works directly.

jspricke (approved):

PR #637, opened 2020-01-28
Release new 1.0.20 version

Gitsaibot:


jspricke:
Some maybe additions to the changelog:
- switch license to GPLv3
- add gradle wrapper
what do you think?

jspricke:
* Gitsaibot <notifications@github.com> [2020-01-28 11:35]:
>Yes I missed that with the license.

can you add it?

>I would left out the gradle wrapper because it doesn't mean anything to the
>"normal user" and he doesn't get anything out of it ;)

Totally fine with me, rather mentioned it for completeness.

PR #680, opened 2020-04-09
improved metadata and screenshots v1

Gitsaibot:


jspricke:
I've renamed the files so we get the best of both PRs, tables screenshots go into the separate directory and phone screenshots overwrite the old files.

jspricke:
@Gitsaibot that's all on purpose, as I tried to explain above.
- We need to keep the old file names for now, so F-Droid overwrites the old data.
- There where not enough new phone pictures so I took one from the old data.
- The tablet data is new so we can give it new names.

jspricke:
> > There where not enough new phone pictures so I took one from the old data.
> 
> Okay, but @PanderMusubi wanted to remove the "phone" frame, now we have 2 pictures with one frame... And why do you name some of them t1. and not p1.-p10. ?

Those are the old file names we want to overwrite (the name actually doesn't matter).
If you provide more screenshots we can replace those with a frame.

> > The tablet data is new so we can give it new names.
> 
> I thought F-Droid didn't delete old directories. As long as we don't change the files, I think that's not a problem, but if we do, which images did F-Droid show ?

I don't get your question, F-Droid should show all (and only) files you can see in this PR.

PR #688, opened 2020-05-06
Drop dependency on kotlin-android-extensions

luk1337:
Since we are only using this extension for a single fragment I think it makes sense to drop it in order to make AOSP port easier.


jspricke:
Compiling works for me, @Gitsaibot can you have a look (as you wrote that code)?
Also, @dschuermann do you have any comments?

dschuermann:
Nope, that's fine and makes sense!

PR #689, opened 2020-05-06
Return true in onNavigationItemSelected()

luk1337:
* That way navigation item gets selected properly on AOSP.


luk1337:
Here's how it behaves without this patch here:
ADB Screenshot from 2020-05-06 17-23-31

Issue #278, opened 2017-12-04
Show correctly STATUS:CANCELLED event


Spheerys:
Hi,

I have the use of the Canceled status for some events on a distant .ics file
The code on the event is : 
```
STATUS:CANCELLED
```
But Etar don't show these events differently than the normal ones.

Could it be possible to show the canceled events striped and eventually lighten ?

Spheerys:
Up to my 2 years old proposition !
I saw there was several issues openned on the same idea :
https://github.com/Etar-Group/Etar-Calendar/issues/71
https://github.com/pimutils/khal/issues/587

Any chance to see this nice improvement implemented ?

Spheerys:
Pleaaaase :)

Kalytis:
Couldn't agree more, it would be nice to edit, or at least show, the status of events. Better integration with Thunderbird !!

MathieuSchopfer:
That would be nice to have.

Calendar Contract expect a `STATUS` column with three possible states: `TENTATIVE`, `CONFIRMED`, and `CANCELED`.

But how should this interact with status-dependent formatting properties (e.g. attendance status) ?

MathieuSchopfer:
I think we can close this issue.

PR #697, opened 2020-05-16
Show canceled events using strikethrough

MathieuSchopfer:
This is a basic implementation of #278.

Screenshot_1589639208
Screenshot_1589642789
Screenshot_1589644345

### DIFF
--- a/src/com/android/calendar/DayView.java
+++ b/src/com/android/calendar/DayView.java
@@ -2800,6 +2800,13 @@ public class DayView extends View implements View.OnCreateContextMenuListener,
                     break;
             }
 
+            if (event.status == Events.STATUS_CANCELED) {
+                // Strike event title if its status is `canceled`
+                paint.setStrikeThruText(true);
+            } else {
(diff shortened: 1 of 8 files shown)


MathieuSchopfer:
Indices would be hard-coded somewhere ? I am going to revert it then. By _rejected events_, you mean events for which you got an invitation, which you refused ?

MathieuSchopfer:
Thanks for testing !

> It seems Re-sorting the Event IDs has negative effects on the month view. Greying out rejected events did not work. Also, the color of the event names changes from white to black in month view(light theme).

It was a stupid mistake on my side. The event `TextPaint` was overridden by the default paint in some situations. So, re-indexing event IDs has -- as expected -- no side effect. (The opposite would mean that one of the `Instances....` or `PROJECTION_..._INDEX` constants would be defined somewhere out of the `Event` class, which is not the purpose of a constant.)

I amended the related commit. Feel free to review again.

jspricke:
Thanks!

PR #700, opened 2020-05-18
Change search provider url

Gitsaibot:


AndroidManifest.xml
```diff
@@ -188,7 +188,7 @@
 
         <provider android:name="com.android.calendar.CalendarRecentSuggestionsProvider"
```
dschuermann, line 189: I think we should also change the provider name to prevent conflicts with AOSP calendar.

Gitsaibot: Do you know how ? In CalendarRecentSuggestionsProvider.java we use the package com.android.calendar and if I change the provider name I have to change the package name. But then I get an file path error.

dschuermann: Ah I see. Yeah, I think I was wrong, since the name is not public, only the authorities.

PR #709, opened 2020-05-28
Move input field for location under the guest field

Gitsaibot:


jspricke:
Can you give some reason for this?

PR #715, opened 2020-06-09
Make use of system theme setting on android 10

Gitsaibot:


res/values-v29/array.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
```
mikeNG, line 1: Should name this file arrays.xml for consistency

Gitsaibot: You are right. Thanks for reviewing

src/com/android/calendar/DynamicTheme.java
```diff
@@ -93,8 +133,15 @@ public static String getPrimaryColor(Context context) {
         return Utils.getSharedPreference(context, COLOR_PREF, TEAL);
     }
 
-    private static String getSuffix(String theme) {
+    private static String getSuffix(String theme, Context context) {
```
jspricke, line 136: API wise this looks a little strange as you get theme twice (once within context). I would propose to drop the theme argument and call getTheme in here.

PR #732, opened 2020-07-07
WIP: removed bolding from agenda

newhinton:


res/values/colors.xml
```diff
@@ -117,6 +117,8 @@
     <color name="agenda_item_declined_color">#66333333</color>
     <!-- Text color of non-declined events in agenda view -->
     <color name="agenda_item_standard_color">#FF333333</color>
+    <!-- Text color of past events in agenda view -->
+    <color name="agenda_item_past_color">#989898</color>
```
jspricke, line 121: This is too similar to the background, making it hard to read.
I would propose to keep the black.

newhinton: I would advise against this. I agree, it is harder to read, but the intention here is to clearly hide everything that is not important. Past events are not important, and should therefore be very different from events that are not in the past. I can try to tweak it, but if we need to increase the contrast we should change the background aswell.

jspricke:
Feel free to propose new colors with more contrast (screenshots would be 
awesome).

res/values/colors.xml
jspricke: Feel free to propose new colors with more contrast (screenshots would be awesome).

newhinton: image

I darkened the text, and made the background just so slightly grey (#F0F0F0)

newhinton: in comparison:

image

jspricke: I vote against it, past events are clearly separated by background color and the red line.

newhinton: We should try to get away from as much dark colours as possible (in the light theme), it quickly gets too heavy and crowded. The lighter foreground does not conflict with the white background of future events, and the dark but not black textcolor moves the focus away to more recent events

newhinton:
I cannot produce the crash on dark modes with the last merge.

PR #734, opened 2020-07-07
Feature/730/bottombar calendartoggle

newhinton:


jspricke:
I think this takes too much space on screen. I would be ok with the drawer as written before.

newhinton:
I disagree. I forgot to mark this, but it is still work in progress (there are still bugs and missing features)

The final version will contain a groupmanager, which will result in the bar not beeing displayed when no groups are set. If screenspace is a problem and this bar takes to much space, it can be disabled. Not only that, but the default will be that it is not shown. That should allow everyone to use the screen to its max. 

Also the drawer is a very bad place for toggleable items, it is meant to provide navigation, not visibilitysettings. Also it's size is even smaller, and calendars would use even more space in there on small screens (albeit, i admit, since its not the "mainview" this is not really a concern, but it still clutters the menu quite heavily, something we dont want)

newhinton:
Ill close this, since it is not required and too much effort to make it work properly for all.

PR #736, opened 2020-07-08
WIP: Overhaul EditEvent Userinterface

newhinton:


res/layout/edit_event.xml
```diff
@@ -70,19 +78,33 @@
...
                 android:layout_marginStart="0dip"
                 android:layout_marginTop="0dip"
-                android:layout_gravity="start|top"/>
+                android:visibility="gone" />
         </FrameLayout>
     </ScrollView>
-</LinearLayout>
+
+
+    <com.google.android.material.floatingactionbutton.FloatingActionButton
```
jspricke, line 99: why this?

res/menu/event_info_title_bar.xml
```diff
@@ -20,14 +20,14 @@
         android:alphabeticShortcut="s"
         android:title="@string/share_label"
         android:icon="@drawable/ic_share_event"
-        app:showAsAction="always" />
+        app:showAsAction="ifRoom" />
```
jspricke, line 23: why this? Dito the two below.

newhinton:
@Gitsaibot could you provide a screenshot how google does it?

I think in the current form, the save&cancel actions take way to much space for it's function

res/menu/event_info_title_bar.xml
newhinton: They are not often used, and create a cluttered ui. We should show only often used and important actions, and hide the rest. Showing 4 or more icons in the topbar is to much and unnessesssary.

res/layout/edit_event.xml
newhinton: See here:
https://github.com/Etar-Group/Etar-Calendar/pull/736#issuecomment-659226086

newhinton:
I added the X button (it may be too big) and coloured the fab aswell

newhinton:
@Gitsaibot why? it hasnt been merged, and i still think that etar could use an overhaul

newhinton:
@Gitsaibot No i did not see your changes.

 I just checked them out, and i have to say, your spacer are waaay better than mine. I also like the icons.
I still think the topbar could be switched to the one i am using, which frees up a bit more space. If you want me to, i can make a pr for it, but this one can be closed i guess

PR #737, opened 2020-07-09
Improved Navigation Drawer

newhinton:


newhinton:
Theming does not work currently for the headerbackground. Are there suggestions on what to do there? A blank screen is somewhat boring, but it needs to be themed at least

newhinton:
Yes, i tried a layered image, but that didnt work, so at first i tried the external lib to get it working. But i think removing it before merge should be doable.

I have read that etar ships with los, and that sounds great! I know, but my goal was to reuse the normal icon anyway, so that we dont have multiple icons. That way, logochanges should be reflected everywhere anyway.

Regarding the color: while it is acceptable to simply have a color, i hoped to get an effect in there, but i wasnt sure what or how to do this, so im open for suggestions.

If we design the logo on top properly (eg. add a dropshadow, a border, etc) the logo should stand out properly regardless of it is. 
Or we remove the icon completely, that is also an option, but i'd prefer to keep it

I have tested it on a device 480x800, and all items are still visible, so it should be fine
<details><summary>Screenshot</summary>
<p>

Screenshot_1594368814

</p>
</details>

newhinton:
@Gitsaibot this is a great idea, it adds a little playfulness. But i am not sure if it works with the redesign of the icon where i redid it as an svg/xml, but i will try!

newhinton:
@Gitsaibot This was not really supposed to happen with the rebase, is this PR still mergeable?

PR #740, opened 2020-07-15
Feature/731/slim dayview

newhinton:


newhinton:
Any news on this? Especially with android 12 etar will have to make changes to it's userinterface, at least to it's widgets

dumblob:
The new "mini" rounding looks very good to me. @jspricke wdyt?

Issue #417, opened 2019-01-09
Widget reactivation


SubSi:
Seems the widget needed some improvements, but are you planning to reactivate it again?

jspricke:
If someone provides a patch to fix these:
https://github.com/Etar-Group/Etar-Calendar/issues?q=is%3Aopen+is%3Aissue+label%3Awidget

bentolor:
Thanks @SubSi for asking that question and thanks for @jspricke for your timely answer.

I understand the rationale behind dropping a functionality in a FOSS project that causes more noise & efforts than resources are available. But I wonder if this is really the situation in this case…

In my humble opinion the calendar widget is _the core feature_ of a calendar program at all. I heavily doubt that in the current form Etar will be able to gain new users and even less motivate contributors to fix these issue. So wouldn't it be more honest to drop/pass the maintainer project as a whole? Maybe someone else is willing to drop in?

On a short skimming through the list of issues mentioned above:

#402 #228 #223 #190 #174 #162 are solely enhancement request
#272 was a bug request, has changed to a documentation request

#385 #373 #374 are the remaining, real bugs. **None** of these bugs seem to cause any severe data loss or massive negative side effect. Two off them could IMHO be at offered a workaround by offering a reload-button on the widget for the time being.

Honestly I find #412 a little overshooting the issue.

Gitsaibot:
> In my humble opinion the calendar widget is the core feature of a calendar program at all.

I don't use the widget at all. Nevertheless, I contributed something to Etar.  So it depends on the users.

> So wouldn't it be more honest to drop/pass the maintainer project as a whole? Maybe someone else is willing to drop in?

Okay and where is he ? I would be glad that someone leads the project in his free time at all...

> Honestly I find #412 a little overshooting the issue.

Overshoothing? That's only a temporary deactivation because with target api 28 the widget doesn't work at all. We waited more than half a year and no one came up with a pull request. I tried it myself but it looks easier than it is. And because I do not use the widget, I've given up. But feel free to make a pull request which solves #373 #374 ;). Then we can activate it again...

ramack:
I was using the widget without any problem but I am not yet on Android O. And I didn't use it intensivley. Most of the time I just open Etar to see the calendar. So I don't see the widget as the most important thing in a calendar app and will be happy without it.

[...]

Issue #374, opened 2018-09-28
Widget doesn't update when new day has come for target SDK 27+


xsoh:
By targeting `targetSdkVersion 27` Android has  [limits background execution](https://developer.android.com/about/versions/oreo/background) and which also limited the broadcasts receiver when the app is in background. 
Etar used `android.intent.action.PROVIDER_CHANGED` to update the date of the widget but now Android will not broadcast this when the app is in background. Only some [broadcasts exceptions](https://developer.android.com/guide/components/broadcast-exceptions) are there.

Issue #373, opened 2018-09-28
Widget doesn't update for new events for target SDK 27+


xsoh:
Widget doesn't update for new events for `targetSdkVersion 27`

marc0w:
Are there any plans to fix it in 2019?

lx4r:
I would love to have a widget for Etar, any updates on this? :)

Issue #272, opened 2017-11-21
Widget doesn't show up on Android 8 Oreo


Whisprin:
The widget of Etar doesn't appear in the add widget dialog of my launcher on Android 8.

I installed Etar from Play Store and Fdroid, version 1.0.12, on my Sony Z3 Compact running Android 8 (Omni Rom, Oreo). Reinstalling and reboots didn't help either.

chipmunk14:
I have the same issue with the widget. Running Android 7.1 (Lineage OS 14.1) on a Moto G Osprey. Tried installing from F-droid and the Play Store and rebooting, but it didn't help. 
Got Etar with widget working on another phone; nexus 5x with Lineage OS and F-droid. All the settings appear to be the same.

sercasbh:
I have same issue.
Android 7.1.2 Resurrection Remix, on Moto G 2014.
Thank you for this app

Gitsaibot:
Without a debug log or something it's hard to determine why the widget doesn't show up. On my phone (7.1.1) it shows up.

Whisprin:
I'm now running an Android 8 stock rom on my phone and the widget shows up. Maybe it has something to do with the custom roms from @chipmunk14 and @sercasbh. I was running a custom rom too, when the issue occurred.
What about you @Gitsaibot, stock or custom?

Gitsaibot:
I use a stock rom. So maybe you are right.

chipmunk14:
Got one phone with Lineage OS 14.1 with working widget and one with Lineage 14.1 where it doesn't show up. @Gitsaibot, where can I find the logs?

binaryplease:
Found the solution to this:
You have to move then app to the internal storage. I had it installed on the SD-card, widget was not available. After moving it to internal storage it is there. Same with other application's widgets. This should be documented somewhere.

totti4ever:
Hey, I'm on Android 9 with my Xperia xz2c and cannot find the widget either. Also I cannot find where to move the app to internal storage, I guess it already is.

So, any ideas how to handle this?

doktorhorst:
Quote from the readme file:
> -Agenda widget.- (disabled until #373 and #374 are fixed. Use Calendar Widget as an alternative.)

PR #746, opened 2020-07-26
Rework widget so we can activate it again

Gitsaibot:
#272 #373 #374 #417


jspricke:
According to https://github.com/Etar-Group/Etar-Calendar/issues/417#issuecomment-663981593 it still doesn't work. Do you have an idea why?

PR #752, opened 2020-08-01
Use primary color for "month background today color" in light theme.

### LINKED ISSUE #727: feature request: change color of current day
Hello,

on my phone the light green color of the current day is hard to notice. This becomes worse when my phone enables night mode and all colors on my display show up slightly red.

I would appercepte an option to change the color of the current day on my own. There does not seem to be such an option.

### LINKED ISSUE #750: [FR] Option to specify color for active day regardless of theme
By, default light theme very pale green with low opacity, which is REALLY hard to notice in light theme, especially if you don't have perfect eyesight.

Ability to specify custom color would be really useful.

Gitsaibot:
Fix #727; Fix #750 The implementation is somewhat different than desired but should improve the situation.

### DIFF
--- a/src/com/android/calendar/month/MonthWeekEventsView.java
+++ b/src/com/android/calendar/month/MonthWeekEventsView.java
@@ -50,6 +50,7 @@ import android.view.MotionEvent;
 import android.view.accessibility.AccessibilityEvent;
 import android.view.accessibility.AccessibilityManager;
 
+import androidx.core.content.ContextCompat;
 import java.util.ArrayList;
 import java.util.Formatter;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


res/values/colors.xml
```diff
@@ -94,7 +94,6 @@
 
     <!-- Background Color of the selected week for the mini calendar on the tablet size -->
     <color name="month_selected_week_bgcolor">#FFDDDDDD</color>
-    <color name="month_today_bgcolor">#25a2ff00</color>
```
jspricke, line 97: Isn't that still used here? https://github.com/Gitsaibot/Etar-Calendar/blob/MonthTodayColor/src/com/android/calendar/month/MonthWeekEventsView.java#L246
Also there is month_today_bgcolor_black and month_today_bgcolor_dark..

Gitsaibot: Yes but only for dark and black theme.

jspricke: Would it make sense to do the same modification for those two themes as well?

jspricke (approved):

Kein:
This didnt make into 1.0.23,. right?

PR #757, opened 2020-08-19
Dont close EditEventFragment on contextswitch

newhinton:


tamas646:
To be honest, for me the current behavior is very annoying.
I would appreciate if this pull request would be merged.

If someone wants to save the event, then (s)he will tap the save button.
If (s)he doesn't do it and the system run out of memory, then it's her/his problem. (the popup will warn her/him anyway)

> Example: Create an event for the latest concert of your favorite band
> You see the new concert, and switch to etar, creating an event. You input the name, but you forgot some detail. Now, if you switch back to your browser, the event is completely lost and you have to input everything again.

I agree with @newhinton. This is exactly why I find the current behavior annoying.

Please think it over.

src/com/android/calendar/event/EditEventFragment.java
```diff
@@ -560,14 +560,15 @@ public void onPause() {
         Activity act = getActivity();
         if (mSaveOnDetach && act != null && !mIsReadOnly && !act.isChangingConfigurations()
                 && mView.prepareForSave()) {
-            mOnDone.setDoneCode(Utils.DONE_SAVE);
-            mOnDone.run();
+            //mOnDone.setDoneCode(Utils.DONE_SAVE);
```
tchernobog, line 563: Do not comment out code, it can always be retrieved by looking at the previous git commit. Just remove these lines.

src/com/android/calendar/event/EditEventFragment.java
```diff
@@ -560,14 +560,15 @@ public void onPause() {
...
+            //mOnDone.run();
+            Toast.makeText(mActivity, ws.xsoh.etar.R.string.unsaved_event, android.widget.Toast.LENGTH_SHORT).show();
         }
         if (act !=null && (Build.VERSION.SDK_INT < 23 ||
-                    ContextCompat.checkSelfPermission(EditEventFragment.this.getActivity(),
+                ContextCompat.checkSelfPermission(EditEventFragment.this.getActivity(),
                         Manifest.permission.READ_CONTACTS) == PackageManager.PERMISSION_GRANTED))
-            act.finish();
-        super.onPause();
+            // act.finish();
```
tchernobog, line 570: Do not comment out, remove.

szaimen:
@newhinton any update here? :)

newhinton:
@szaimen I have switched to a different calendar quite a while a go, but i can take a look if it is still possible to use this code in the current master!

tamas646:
@newhinton I would be grateful for your effort too.

szaimen:
> @szaimen I have switched to a different calendar quite a while a go, but i can take a look if it is still possible to use this code in the current master!

Yes, would be great! Thanks a lot! :)

newhinton:
I removed the comments, maybe it gets merged now!

szaimen:
@Gitsaibot can you please review this? Thanks a lot! :)

newhinton:
Business Calendar 2 does it this way. If their edit-view is opened from the backstack it will show the open editor, but if launched from the app-launcher, the backstack is discarded.

jspricke:
I think the correct way would be to store the data in the instance state as discussed here: https://developer.android.com/guide/components/activities/activity-lifecycle

newhinton:
> I think the correct way would be to store the data in the instance state as discussed here

Isn't that exactly what android does automatically?

Etar just forcefully discards the state in the onPause()-call, interrupting android's own lifecycle.
This PR just let's android handle the activity on its own.

PR #759, opened 2020-08-19
go back to default view instead of quitting (except dayview&agendaview)

newhinton:


jspricke:
Yes I can reproduce it with this one commit on top of master. Things I did:
- fresh install of Etar
- open an event
- press back (3 times)
- open Etar
- open an event

jspricke:
Probably.

PR #760, opened 2020-08-19
overhauled about-fragment to make it less cramped

newhinton:


newhinton:
yup, that is expected. I created a landscape version of the about xml, because just stacking everything on top of each other does not look good. So i basically created two adjacent "stacks" that display the information, and therefore a new file was required

newhinton:
Screenshot_1643192192
Screenshot_1643192323


@Gitsaibot i have added screenshots and rebased my branch. The huge statusbar is an artifact of the emulator

newhinton:
@Gitsaibot This should resolve the conflicts

PR #768, opened 2020-08-31
Settings: add a button to add remote EteSync accounts.

tasn:
EteSync is a fully open-source (clients and server), end-to-end encrypted,
and privacy respecting sync for your contacts, calendars and tasks.

This change was kept as close as possible to the existing CalDAV/DAVx5 button
so it made it possible for me to easily add translations for all languages that
already had the CalDAV button translated.

### DIFF
--- a/res/values-ca/strings.xml
+++ b/res/values-ca/strings.xml
@@ -311,6 +311,7 @@
     <string name="preferences_list_add_offline_title">Afegeix un calendari fora de línia</string>
     <string name="preferences_list_add_offline">Afegeix un calendari fora de línia</string>
     <string name="preferences_list_add_remote">Afegeix un calendari remot (CalDAV)</string>
+    <string name="preferences_list_add_remote_etesync">Afegeix un calendari remot (EteSync)</string>
     <string name="preferences_list_general">Configuració general</string>
     <string name="preferences_list_calendar_summary_invisible">No es mostren els esdeveniments</string>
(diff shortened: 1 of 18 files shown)


tasn:
EteSync is not CalDAV, it's another open protocol, so I think EteSync should remain `(EteSync)`.
As for DAVx5 - I know you didn't ask me but maybe `(CalDAV DAVx5)` as in a space instead of a dash?

Gitsaibot (approved):

dschuermann:
The thinking behind the CalDAV button was the following: There are users who discover Etar, but don't know how to add remote calendars. This questions sometimes pop up on the issue tracker.

I would propose to use short simple button labels:
```
    <string name="preferences_list_add_remote">Add CalDAV calendar</string>
    <string name="preferences_list_add_remote_etesync">Add EteSync calendar</string>
```

tasn:
Would you like me to change the string? Though FYI, it's already translated to quite a few languages, so breaking the string means we will lose all of these translations.

dschuermann:
I would propose to change the strings. @Gitsaibot what's your stance on this?

tasn:
Would you like it done as part of this PR, or a later change that changes the strings for both DAVx and EteSync?

tasn:
OK, this means I'm going to remove all of the translations for the EteSync and DAVx5 strings and just leave English. If you are OK with that, let me know, and I'll update the PR.

dschuermann:
ACK

tasn:
Done.

PR #780, opened 2020-09-21
[Draft] Implement agendawidget

newhinton:
This is a testimplementation for an simple agendawidget. I based it on the existing widgetcode, which i extended to easily add new/different widgets.


newhinton:
image

newhinton:
Theming "works" (i need quite a lot of different drawables, because i cant set colors dynamically, gladly i only need the single colors and not combinations of colors and themes)
I needed to make some stuff public from DynamicTheme for that. Loading is a bit buggy, and clicking on a single entry does not open the corresponding event but the App in general

newhinton:
@Gitsaibot @jspricke  So i got it to work stable, so the bare minimum has been achieved. However, i had to touch many parts of the old widget code, and i would suggest that i furter alter it, to create a "generic" widget, that can be used in other widgets. 

For example, one "problem" with the existing code is, that it adds the first element twice, presumably to show it as an header. That does not work with my implementation, therefore that needs to be removed. In my dreamimplementation, we have a widgetservice that only supplies the logic to load data and update the widget, while the code associated with displaying lists and creating the actual widget needs to be implemented by smaller widgets. 

Are there any plans that would interfere with those changes, or can i continue to strip down the existing code and make it more generic?

tonya1545b:
Any update on this?

newhinton:
Closed in favor of #1078

PR #782, opened 2020-09-26
Don't show week number in miniMonth

### LINKED ISSUE #773: Month-View in sidepane wrong when showing weeknumbers
I just realized when in single day view with side pane open, the month view is wrong. The selected day is not the shown.
Only happens when the weknumbers are shown.

Ways to reproduce:
- Enable weeknumbers in settings
- Go to day view
- select day in little month-display
- The day next to the touched one shows up
- Disable weeknumbers to get back to normal

I think it is mostly a problem with displaying stuff since the head with weekday letters is not shifted with the rest.
Might help to introduce a new first column in head to display over weeknumber?

Screenshot_20200910-214948_1

Gitsaibot:
Fix #773

- Screenshot [before](https://user-images.githubusercontent.com/15521729/94351158-6b297a80-0056-11eb-8e17-67214aa60130.png)

- Screenshot [after](https://user-images.githubusercontent.com/15521729/94351189-a62bae00-0056-11eb-874b-7df159bacb0d.png)

### DIFF
--- a/src/com/android/calendar/month/MonthByWeekAdapter.java
+++ b/src/com/android/calendar/month/MonthByWeekAdapter.java
@@ -305,7 +305,11 @@ public class MonthByWeekAdapter extends SimpleWeeksAdapter {
     @Override
     protected void refresh() {
         mFirstDayOfWeek = Utils.getFirstDayOfWeek(mContext);
-        mShowWeekNumber = Utils.getShowWeekNumber(mContext);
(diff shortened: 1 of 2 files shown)


jspricke:
This is fine with me, though it would be nice if we could keep the week numbers. Did you look into adding an extra column head instead?

jspricke (approved):

PR #786, opened 2020-10-04
View binding

Gitsaibot:
Make use of view Binding in activity's. See https://developer.android.com/topic/libraries/view-binding

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -201,6 +201,10 @@ android {
         }
     }
 
+    buildFeatures {
+        viewBinding true
+    }
+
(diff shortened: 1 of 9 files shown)


jspricke (approved):

mikeNG:
For reference, this breaks the AOSP compiled app since view binding is not supported in AOSP.

jspricke:
@mikeNG thanks for the heads up!
Is there some way to ifdef it?
Do you have more information why?

Can you open a separate issue so we can track this?

PR #787, opened 2020-10-07
Added possibility to choose black theme for night mode

### LINKED ISSUE #419: [Feature Request] Auto Night Mode
Functionality desired: automatically switch between light and dark/night themes depending on time of day.

Something akin to QKSMS's setting (Screenshot Attached):
screenshot_20190109-174446_qksms

Gitsaibot:
Fix #419

### DIFF
--- a/src/com/android/calendar/DynamicTheme.java
+++ b/src/com/android/calendar/DynamicTheme.java
@@ -52,11 +52,11 @@ public class DynamicTheme {
 
     private static int getSelectedTheme(Activity activity) {
         String theme = getTheme(activity) + getPrimaryColor(activity);
-        boolean pureblack = Utils.getSharedPreference(activity, PURE_BLACK_NIGHT_MODE, false  );
+        boolean pureBlack = Utils.getSharedPreference(activity, PURE_BLACK_NIGHT_MODE, false);
         switch (theme) {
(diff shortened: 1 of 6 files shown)


src/com/android/calendar/DynamicTheme.java
```diff
@@ -51,10 +52,15 @@ private static String getTheme(Context context) {
 
     private static int getSelectedTheme(Activity activity) {
         String theme = getTheme(activity) + getPrimaryColor(activity);
+        boolean pureblack = Utils.getSharedPreference(activity, PURE_BLACK_NIGHT_MODE, false  );
```
jspricke, line 55: nit: remove spaces after false.

jspricke:
I don't find the option in settings (German locale)

jspricke:
Thanks for the hint, can we show it even if the phone is not in night mode?

jspricke:
What about just disabling the screen refresh in light mode? I think options should be independent from the current mode, I want to enable it even if I'm currently in light mode.

jspricke:
Works for me, thanks for the update :).
It may make sense to put it directly under the design option as it's closer from the setting, what do you think?

jspricke (approved):

Issue #201, opened 2017-03-23
Handling the event STATUS property


patricktokeeffe:
As noted in [this comment](https://github.com/Etar-Group/Etar-Calendar/issues/71#issuecomment-258080690), event statuses are not always handled correctly. I'm using Etar (1.0.10) + Davdroid (1.4.0.3-ose) + Owncloud (8.x), and I observe odd handling of the STATUS property:

1. Create event using third-party client (in my case, Lightning/Thunderbird) 
2. Edit event in Etar
3. Observe status is now "Tentative"

The problem appears to hinge on the original event not having any status specified:

![2017-03-23 13_52_18-new issue etar-group_etar-calendar - chromium](https://cloud.githubusercontent.com/assets/1348834/24269836/fef5ef62-0fcf-11e7-9f02-5fdde067258e.png)

````
BEGIN:VEVENT
CREATED:20170323T202138Z
LAST-MODIFIED:20170323T202150Z
DTSTAMP:20170323T202150Z
UID:c2606d80-deab-4c5f-8a89-71ef56445976
SUMMARY:test event
DTSTART;TZID=America/Los_Angeles:20170330T140000
DTEND;TZID=America/Los_Angeles:20170330T150000
TRANSP:OPAQUE
CLASS:PUBLIC
END:VEVENT
````

After editing the event with Etar, there is a new property `STATUS` which has the (annoying) default value of `TENTATIVE`.

> Side note: a second new property `SEQUENCE` is also added, and the properties `LAST-MODIFIED` and `TRANSP` are both removed.

````
BEGIN:VEVENT
CREATED:20170323T202138Z
DTSTAMP:20170323T202520Z
UID:c2606d80-deab-4c5f-8a89-71ef56445976
SUMMARY:test event
STATUS:TENTATIVE
DTSTART;TZID=America/Los_Angeles:20170330T140000
DTEND;TZID=America/Los_Angeles:20170330T160000
SEQUENCE:1
CLASS:PUBLIC
END:VEVENT
````

The iCal spec does not indicate which, if any, value is the correct default for `STATUS` property. I believe the correct behavior on Etar's part is to ignore this property if it doesn't exist (not to create it and set it to an arbitrary value).

Inrumpo:
I'm experiencing a similar issue: Creating an event in Etar on Android 11 (CalyxOS) will display it as "tentative" in Thunderbird. I then have to go to Tunderbird and set all those events to "confirmed" manually.

[...]

Issue #71, opened 2016-07-13
Ability to set status of events


madduck:
Please allow [the STATUS](https://tools.ietf.org/html/rfc5545#section-3.8.1.11) of events to be changed through Etar (confirmed, cancelled, tentative).

Thanks!

hibaer:
Hi, I have to follow up on this point. I'm using Etar (1.0.7) in combination with a nextcloud server (10.0 stable). Sometimes it presumably happens, that the status is not transferred correctly, i.e., Etar says that the status is "CONFIRMED", while owncloud says "TENTATIVE". This seems to happen randomly, I could not deliberately reproduce it. Has somebody observed similar issues? The error may also reside on the nextcloud server.

nolash:
Is this being worked on?

Inrumpo:
This feature request is now 6 years old. 🎂
I still hope to see that feature sometime in the future.

PR #800, opened 2020-11-01
Add ability to edit event status (#71, #201)

MathieuSchopfer:
- Add a spinner to the `Edit Event View` to select the event status,
   one of `CONFIRMED`, `TENTATIVE`, or `CANCELLED`.

 - Set default status in `Calendar Event Model` to `CONFIRMED`.

Screenshot_1604267796
Screenshot_1604267800


res/values/arrays.xml
```diff
@@ -241,13 +241,20 @@
...
         <item>2</item>  <!-- Tentative -->
     </integer-array>
 
+    <!-- CalendarContract.EventsColumns#STATUS -->
+    <string-array name="event_status">
+        <item>Tentative</item> <!-- 0 -->
+        <item>Confirmed</item> <!-- 1 -->
+        <item>Cancelled</item> <!-- 2 -->
+    </string-array>
+
```
Gitsaibot, line 257: This should be linked to "strings" for translation(https://github.com/WeblateOrg/weblate/issues/4269). Same as I did with string array visibility in line 258...

MathieuSchopfer: Ok, that is why you did it this way ... It is a bit counter-intuitive though to have to set the property `translatable` to `false`. It led me to think that without this property, the strings to translate would be extracted directly from this array.
Anyway, I pushed the fix.

MathieuSchopfer:
> Thank you very much for the pull request. This will have to wait until my changes to EditEvent are applied.

No problem, I'll rebase when you're done.

By the way, I only integrated the change in layout `edit_event_1.xml`. Should we also integrate it somewhere else ? I do not know which layouts are still in use (and where they're active in the app).

res/values/arrays.xml
Gitsaibot: Yes, I hope the bug will be fixed sometime ;).

jspricke (changes requested):

src/com/android/calendar/event/EditEventFragment.java
```diff
@@ -642,8 +642,7 @@ public void onColorSelected(int color) {
         long end = -1;
     }
 
-    // TODO turn this into a helper function in EditEventHelper for building the
-    // model
+    // TODO turn this into a helper function in EditEventHelper for building the model
```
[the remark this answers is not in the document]
MathieuSchopfer: I reverted that.

MathieuSchopfer:
There is a bug though. When I create a cancelled event, it is struck through in the `AgendaFragment`. (That is what it is expected to look like, see screenshot below.) But then, if I edit this event and change its status to let say `Confirmed`, the `AgendaFragment` does not refresh the view, and the event is still struck through. Do you know if it possible to refresh the event painter in `AgendaFragment` ? (preferably to refresh only the painter of the event that was just edited)

Screenshot_1606771490

> Can you rebase and put the status spinner under the visibility spinner please.

I did that:

Screenshot_1606770929

res/layout/edit_event_all.xml
```diff
@@ -627,6 +626,40 @@
...
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_marginStart="12dp"
+            android:layout_marginTop="20dp"
+            android:layout_marginBottom="20dp"
+            android:src="@drawable/ic_outline_verified"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toBottomOf="@+id/view8"
+            app:layout_constraintBottom_toBottomOf="parent" />
+        <Spinner
```
Gitsaibot, line 648: Please add a line break above "Spinner" for consistency.

res/layout/edit_event_all.xml
```diff
@@ -627,6 +626,40 @@
...
+
+        <!-- Event status -->
+        <ImageView
+            android:id="@+id/event_status_icon"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_marginStart="12dp"
+            android:layout_marginTop="20dp"
+            android:layout_marginBottom="20dp"
+            android:src="@drawable/ic_outline_verified"
```
Gitsaibot, line 644: Get an compile error because of this missing resource.

MathieuSchopfer: Oups, sorry, I failed to include this file in the commit.

MathieuSchopfer:
> If I set the status to CANCELLED, the event is no longer visible in all views !? So on my cell phone they are not crossed out but invisible. Any idea why ?

No, it should appear crossed out. I can reproduce this bug. It seems the event is somehow destroyed. Strangely, if you

- *create* a cancelled event, it appears correctly.
- change the status of any event to tentative or confirmed, the event is not destroyed.

I am trying to investigate that. But it goes much beyond my knowledge of the app and of Android framework I am afraid ... If anyone with more insights can make an assumption, that would nice.

akki42:
Sorry to "bump" this topic, but there seems to have been no updates to the STATUS-issues (this commit as well as https://github.com/Etar-Group/Etar-Calendar/issues/71 and https://github.com/Etar-Group/Etar-Calendar/issues/201) since Mathieu's proposals and this discussion in December last year.
Is anybody still looking at this? Would so love to have these features in Etar. ;-)

dilyanpalauzov:
Apart from TENTATIVE, CONFIRMED and CANCEELLED it shall be possible to set no status.  The screenshot above does not suggest so.

dilyanpalauzov:
Apart from TENTATIVE, CONFIRMED and CANCEELLED it shall be possibe to set no status.  The screenshot above does not suggeat so.

Thannoy:
> > If I set the status to CANCELLED, the event is no longer visible in all views !? So on my cell phone they are not crossed out but invisible. Any idea why ?
> 
> No, it should appear crossed out. I can reproduce this bug. It seems the event is somehow destroyed, but I exported my calendar to ICS and checked : the event is correctly edited with the `CANCELLED` status.
> 
> Strangely, if you
> 
> * _create_ a cancelled event, it appears correctly.
> * change the status of any event to tentative or confirmed, the event is not destroyed.
> 
> I am trying to investigate that. But it goes much beyond my knowledge of the app and of Android framework I am afraid ... If anyone with more insights can make an assumption, that would nice.

You might be interested in issue #1432. Especially, syncing calendars from another calendar app can restore deleted events

akki42:
> > create a cancelled event, it appears correctly.
> 
> But when you change something like the date the event disappears. At the moment I also have no idea why this is happening. This morning all " cancelled" appointments were suddenly visible. But they disappear immediately if I change something like the date. Totally strange...

Coming back to this rather stale issue:
I don't know if you were testing on a synced calendar; if so, this might be related to a "feature" of ical4android / DAVx5, see https://github.com/bitfireAT/davx5-ose/discussions/1132#discussioncomment-11284996.

PR #802, opened 2020-11-04
New EditEvent Layout

### LINKED ISSUE #507: No custom color select for events
Hi,
on Samsung Tab S3 there is no color select for events, as on my Samsung Note 4. Both use Version 1.0.16.

### LINKED ISSUE #742: Done Button Poorly Responsive
I often need to press the Done button to save an event 3 times in a row to get it to work.

This is NOT true in any other apps. You might think something is wrong with my touch screen at that place but in all my other apps that area of the screen works fine.

Is it possible the Done button touch field is smaller than the visual button?

Does anyone else have this issue?

### LINKED ISSUE #801: Feature: New EditEvent design
**New design build with ConstraintLayout**

EditEvent

Feedback is welcome ;)

Gitsaibot:
First Step to a new edit event layout based on ConstraintLayout. Fix #801, Fix #742,  Fix #507


jspricke:
Some thoughts on this:
- The X/Done buttons are inconsistent and rather big, not sure how to improve.
- The activation area for some fields is off (title should be all the way to the right, times are shifted to the left).
- Vertical align the X/Done buttons.
- When editing an Event there is a color plate next to the calendar, but it seem to have no function.

Please clean up the commits.

jspricke (approved):
Thanks for the update, please remove the "Add save button" commit and it's revert from the PR and it should be fine.

PR #808, opened 2020-11-19
Use constraintLayout for about fragment

Gitsaibot:


jspricke (changes requested):
The authors and license buttons no longer work.

jspricke:
Please make the text button work again.

PR #825, opened 2020-12-24
Remove divider on top and focus on day in agendaListView

Gitsaibot:
Before | After
------------ | -------------
Ohne_Tag|Mit_Tag|
Agenda_Divider|Ohne_Divider

### DIFF
--- a/src/com/android/calendar/agenda/AgendaFragment.java
+++ b/src/com/android/calendar/agenda/AgendaFragment.java
@@ -179,7 +179,7 @@ public class AgendaFragment extends Fragment implements CalendarController.Event
             // Set scroll listener so that the date on the ActionBar can be set while
             // the user scrolls the view
             lv.setOnScrollListener(this);
-            lv.setHeaderSeparator(getResources().getColor(R.color.agenda_list_separator_color), 1);
+            //lv.setHeaderSeparator(getResources().getColor(R.color.agenda_list_separator_color), 1);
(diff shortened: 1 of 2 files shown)


src/com/android/calendar/agenda/AgendaFragment.java
```diff
@@ -179,7 +179,7 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
             // Set scroll listener so that the date on the ActionBar can be set while
             // the user scrolls the view
             lv.setOnScrollListener(this);
-            lv.setHeaderSeparator(getResources().getColor(R.color.agenda_list_separator_color), 1);
+            //lv.setHeaderSeparator(getResources().getColor(R.color.agenda_list_separator_color), 1);
```
jspricke, line 182: nit: please remove commented code.

jspricke (approved):

src/com/android/calendar/agenda/AgendaFragment.java
Gitsaibot: I left it in on purpose because I'm not sure if you can use it later.

jspricke: You could get it from the git history in that case.

Gitsaibot: True, removed it

PR #834, opened 2021-01-03
Etar: Support Go devices without widgets permission

AdrianDC:
* Add widget safeguards against unsupported widgets
 * Validate the feature support only once to avoid recurrence

 * Based upon the following commit from packages/apps/Email:
   Commit: AOSP/Email - Added WidgetService#isWidgetSupported
           to check if widgets are supported on the device and
           avoid calling AppWidgetManager#getInstance if not
           supported to avoid NullPointerException.
   Change-Id: I9fff575684e935a96ffe3e3d732a854b9cec461c

LineageOS Gerrit : https://review.lineageos.org/c/LineageOS/android_packages_apps_Etar/+/299989


Gitsaibot (approved):
Looks good! Not tested because no Go device at hand.

PR #894, opened 2021-04-22
Enable lights for "Etar" channel notification by default

luk1337:


luk1337:
* this will only affect new installations *

Gitsaibot (approved):

PR #895, opened 2021-04-28
Use primary color in statusbar

newhinton:
This is part of my ui-overhaul of etar. It removes unnecessary color-differences that do not convey any meaning and are therefore not useful


newhinton:
The effect already works on sdk 21, so this should not be an issue

newhinton:
@Gitsaibot This is because the attribute `android:statusBarColor` only exists in api 21 and later. We can only color the status bar in later versions, and so the new themes file is required.

Also: we cant remove any of the other attributes, because we can only override CalendarAppThemeLight and that has the green color.

newhinton:
Ah that is what you mean. I originally forked etar when min sdk was 19, so that is a leftover. I will merge the two files shortly!

Gitsaibot (approved):

PR #935, opened 2021-06-05
Home time zone setting grayed out if not enabled

Gitsaibot:
Before | After
------------ | -------------
Before | Pref_grey_out

### DIFF
--- a/res/values/themes.xml
+++ b/res/values/themes.xml
@@ -47,10 +47,7 @@
         <item name="colorAccent">@color/colorAccent</item>
         <item name="light_dark">@android:color/white</item>
         <item name="android:windowBackground">@color/bg_dark</item>
-        <item name="android:textColorPrimary">@color/fg_dark_primary</item>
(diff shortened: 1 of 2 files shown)


res/values/themes.xml
```diff
@@ -47,10 +47,7 @@
         <item name="colorAccent">@color/colorAccent</item>
         <item name="light_dark">@android:color/white</item>
         <item name="android:windowBackground">@color/bg_dark</item>
-        <item name="android:textColorPrimary">@color/fg_dark_primary</item>
```
jspricke, line 50: why can we delete those?

jspricke:
Would it make sense to use a switch button instead? (just an idea)

res/values/themes.xml
Gitsaibot: It is a remnant of my very first dark theme customization. As far as I can see is no longer used anywhere.

jspricke (approved):

PR #941, opened 2021-06-19
Refresh view only when themePref "system" is selected

Gitsaibot:


jspricke:
This looks fine in general but sounds a bit counter intuitive, did you test all combinations of before/after?
I may have time tonight to test.

src/com/android/calendar/settings/GeneralPreferences.kt
```diff
@@ -292,8 +292,10 @@ class GeneralPreferences : PreferenceFragmentCompat(),
                 a.recreate()
             }
             KEY_PURE_BLACK_NIGHT_MODE -> {
-                Utils.sendUpdateWidgetIntent(a)
-                a.recreate()
+                if (themePref.value == "system") {
```
jspricke, line 295:
```suggestion
                if (DynamicTheme.isSystemInDarkTheme(a)) {
```
Not sure if this is better, just a proposal. I tried to reproduce the flickering but didn't see it. As it is for the widget, feel free to do what you think fits best.

jspricke (approved):

jspricke:
please revert the merge.

jspricke (changes requested):
.

src/com/android/calendar/settings/GeneralPreferences.kt
```diff
@@ -292,8 +292,10 @@ class GeneralPreferences : PreferenceFragmentCompat(),
                 a.recreate()
             }
             KEY_PURE_BLACK_NIGHT_MODE -> {
-                Utils.sendUpdateWidgetIntent(a)
-                a.recreate()
+                if (themePref.value == "system" && DynamicTheme.isSystemInDarkTheme(a)) {
```
jspricke, line 295: `&&` sounds strange, if it worked without my suggestion before, why would it work now?
Also, my suggestion was based on how we use pure black at other places.

jspricke (changes requested):

src/com/android/calendar/settings/GeneralPreferences.kt
Gitsaibot: Good catch, we need both. Otherwise it flickers in the light theme when system is selected.

src/com/android/calendar/settings/GeneralPreferences.kt
Gitsaibot: Good catch, we need both. Otherwise it flickers in the light theme when system is selected.

Gitsaibot: It is only about when the "pureblack" switch refreshes the settings view. This should only happen when system is preselected and a dark phone theme is selected.

jspricke (approved):

PR #946, opened 2021-06-28
Avoid text overlapping in start, end date fields

### LINKED ISSUE #897: Day and time selectors overlap when a big font size is selected
Version: 1.0.25

When I select the biggest font size (in android screen settings), I see that the day and time selector are together, both are clickable. But...

imagen

Further down in the same view

imagen

I saw that #851 maybe solve the second, but not the first one

04cfb1ed:
Fixes #897

### DIFF
--- a/res/layout/edit_event_all.xml
+++ b/res/layout/edit_event_all.xml
@@ -194,79 +194,66 @@
             app:layout_constraintStart_toStartOf="parent"
             app:layout_constraintTop_toBottomOf="@+id/view1" />
 
-        <LinearLayout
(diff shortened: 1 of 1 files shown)


04cfb1ed:
I tried to use ConstraintLayout at the beginning, but it didn't work, so I took as an example the LinearLayout of the attendees and the organizer. Trying again today, I found the error (some missing params/constraints). I will upload the commit removing the layout in one minute.
As a side effect, the hour won't be wrapped, only the date

Gitsaibot (approved):

PR #995, opened 2021-09-11
Git ignore some .idea/* files

mueller-ma:
Git wanted to commit a lot of files under the .idea directory. This PR
contains the latest `.gitignore` from https://github.com/github/gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,12 @@
+# Source: https://github.com/github/gitignore/blob/master/Android.gitignore
+
 # Built application files
 *.apk
+*.aar
 *.ap_
+*.aab
 
-# Files for the Dalvik VM
+# Files for the ART/Dalvik VM
 *.dex
 
 # Java class files
@@ -11,6 +15,9 @@
 # Generated files
 bin/
 gen/
+out/
+#  Uncomment the following line in case you need and you don't have the release build type files in your app
+# release/
 
 # Gradle files
 .gradle/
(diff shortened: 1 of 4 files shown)


jspricke:
This has nothing to do with Etar but rather with your local setup, I will reject this.
`.idea/*` are probably files from your editor, you can use add them to your `.gitignore` in your home folder or simply add a `.gitignore` in your local git clone. Also see: https://git-scm.com/docs/gitignore
Btw. git does not "want" to commit new files by default, you need to add them manually with `git add`. Maybe you used something like `git add .` and accidentally added all files? The better workflow is to use `git status` and manually select those files you want to commit or use `git add -i`.

mueller-ma:
Honestly I disagree in two things here:

> .idea/* are probably files from your editor

That's correct, but in comparison to other frameworks, there's only one broadly used IDE for native Android apps: Android Studio. IMO in this case it's worth to add these files to the `.gitignore` of the repo as every contributor has to add them to their local `.gitignore`.

> Maybe you used something like git add . and accidentally added all files? The better workflow is to use git status and manually select those files you want to commit or use git add -i.

You're right, I ran `git add .` before. IMO it's not practical to `git add` every changed file manually, especially since `git status` shows a lot of files after starting Android Studio:
````bash
$ git status | wc -l
84
````

jspricke:
* mueller-ma ***@***.***> [2021-09-12 08:13]:
>You're right, I ran `git add .` before. IMO it's not practical to `git add` every changed file manually, especially since `git status` shows a lot of files after starting Android Studio:
>````bash
>$ git status | wc -l
>84
>````

I just tried Android Studio on a fresh clone and git status only shows 
`.idea`.
If you want to commit all your changes at once use `git commit -a`, `git 
add .` is the wrong command almost always.

jspricke:
* Gitsaibot ***@***.***> [2021-09-13 12:17]:
>For me it would be also ok if we exclude the whole .idea folder per gitignore. This would be a single additional entry in gitignore ;)

That would be ok with me.

mueller-ma:
I opened #999

PR #1035, opened 2021-11-07
Skip Reminders option is not showing selected preference on the popup

kavitamp:
`Skip Reminders` option has two different values, namely `Only if declined` & `If declined or not responded`. Even if we select the second option that is `If declined or not responded`, it always shows the first option(Only if declined) as the selected option.

### Steps:
1. Open `Settings` from the side menu
2. Tap on `General settings`
3. Tap on the last option `Skip Reminders`
4. Select `If declined or not responded` from the popup
5. Again tap on the `Skip Reminders` option
6. It always shows `Only if declined` as the selected option

### Screen Recording

https://user-images.githubusercontent.com/20037228/140657322-ab4831e7-d293-4160-a677-c84a5e99bc19.mp4

### DIFF
--- a/src/com/android/calendar/settings/GeneralPreferences.kt
+++ b/src/com/android/calendar/settings/GeneralPreferences.kt
@@ -34,6 +34,7 @@ import android.provider.CalendarContract.CalendarCache
 import android.provider.SearchRecentSuggestions
 import android.provider.Settings
 import android.text.TextUtils
+import android.util.Log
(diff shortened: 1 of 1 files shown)


kavitamp:
@Gitsaibot 
Ohh I didn't notice your PR & already raised bug. The code fixes done in both PRs are almost same.
I think you can merge your changes, I will discard mine.
Thanks

PR #1038, opened 2021-11-16
Follow system palette on devices supporting Material You

zhaofengli:
This PR updates the target API to 31 and implements wallpaper-based dynamic color scheme on Android 12 devices supporting [Material You](https://material.io/blog/announcing-material-you). On supported devices, the app will use the new `monet` color scheme which follows the system palette, replacing the existing color selection functionality. The result is a personalized look that integrates with the rest of the system:

As of Android 12.0, the color extraction algorithm is closed-source and not part of AOSP, though it is [expected to be added](https://www.xda-developers.com/android-12-1-monet-open-source) in the upcoming release. In order to not take functionality away from users running an AOSP-based ROM, we detect the presence of the static AOSP palette and enable the old theming feature.

To test on an emulator, you need the Android 12.1 / Sv2 image:

- Android 12.1 / Sv2 AVD image: Material You enabled
- Android 12.0 / S AVD image (non-Google Play): No Material You functionality, color picker visible
  - The Google Play variants have the Pixel config overlay applied and the static blue palette will be used.
- Android 11.0 AVD image: No Material You functionality, color picker visible

I have tested this on a Pixel 6 running AOSP with [kdrag0n's color extractor implementation](https://gist.github.com/zhaofengli/232f5a3d33113871ad61491629886084).


zhaofengli:
I just checked my changes to the manifest once more, and flipped the ones related to widget update to "android:exported=false". The other ones should be [necessary](https://developer.android.com/about/versions/12/reference/common-intents-31) for other apps to view/create/edit calendar events, and to open links and calendar event files.

src/com/android/calendar/Utils.java
```diff
@@ -204,6 +204,36 @@ public static boolean isOreoOrLater() {
         return Build.VERSION.SDK_INT >= Build.VERSION_CODES.O;
     }
 
+    /**
+     * Returns whether the system supports Material You.
+     *
+     * As of Android 12.0, Material You is only available on some devices (Pixel, select Samsung
```
Gitsaibot, line 210: nitpick: selected

zhaofengli: No, it's not a typo. Here it means "specific" (as opposed to "specified"/"selected").

zhaofengli: Actually, both are technically [correct](https://english.stackexchange.com/questions/57905/is-there-a-difference-between-select-and-selected), but still I think "select" sounds more natural :shrug:

Gitsaibot: I'm not a native English speaker, so we'd better leave it that way ;)

Gitsaibot (approved):

Issue #353, opened 2018-08-03
Year 2038 problem (32 bit unix time rollover)


earboxer:
I can't navigate past january 2038.
I see the week of january 17-23, but flipping to the next page goes to december 1969.

In addition, the date selector for adding an event won't go past december 2036.

https://en.m.wikipedia.org/wiki/Year_2038_problem

Switching to 64 bit unix time will give us 200 billion extra years to come up with a better solution.

redfast00:
Another bug because of 32-bit timestamps:  I sync birthdays with DAVx5. The birthdays don't have a birth year, since I don't have that information for everyone. Birthdays without year default to being yearly events, starting in 1604. Now all my birthdays are shifted by a few months.

@Gitsaibot I think this is now a bug instead of an enhancement, because this breaks a real-world usecase? Could you please tag it as such?

redfast00:
(With 64-bit timestamps that could also be negative, this problem would be solved)

PR #1047, opened 2021-12-01
Adapt to Android 12 Time API changes

chirayudesai:


src/com/android/calendar/event/EditEventView.java
```diff
@@ -1549,9 +1546,9 @@ private void updateHomeTime() {
                             mActivity, mF, millisStart, millisStart, flags, tz).toString());
 
             // Make any adjustments needed for the end times
-            if (isDSTEnd != isDSTStart) {
+            if (false != false) {
```
chirayudesai, line 1549: Doesn't look right, it's part of "Remove isDst"

chirayudesai:
This is current to 319634: Remove isDst references and always return false | https://review.lineageos.org/c/LineageOS/android_packages_apps_Etar/+/319634 Patch Set 2 '724dc1ea663faab84131b37724a50f36a450d754'

.gitmodules
```diff
@@ -10,7 +10,7 @@
 	ignore = untracked
 [submodule "external/calendar"]
 	path = external/calendar
-	url = https://github.com/Etar-Group/android_frameworks_opt_calendar.git
+	url = https://github.com/mikeNG/android_frameworks_opt_calendar.git
```
jspricke, line 13: can you drop this change? We would like to keep our own version. Feel free to send patches as PRs to that repo.

chirayudesai: Yes I intend to drop it before merging, it's only temporary to make it easier to test this, build will fail otherwise.

https://github.com/Etar-Group/android_frameworks_opt_calendar/pull/1

chirayudesai (commented):
@mikeNG found these, which is what caused the issue.

chirayudesai:
> With this patch monthview looks like this: 

should be fixed now.

The issue:
`weekDay` was changed to `getWeekDay`, and `monthDay` was changed to `getDay` - there was a mixup.

Mike and I both went through the entire change independently to find and fix any remaining mis-matches, and this seems all.

I think the main concern at this point would be the all `isDst` changes

chirayudesai:
Ok fixed for real this time.

image

Issue:
`time.monthDay++` was replaced by `time.getDay() + 1`, missing a call to `setDay()`

src/com/android/calendar/month/SimpleWeekView.java
```diff
@@ -285,29 +285,30 @@ public void setWeekParams(HashMap<String, Integer> params, String tz) {
...
                 mFocusDay[i] = false;
             }
-            if (time.year == today.year && time.yearDay == today.yearDay) {
+            if (time.getYear() == today.getYear() && time.getYearDay() == today.getYearDay()) {
                 mHasToday = true;
                 mToday = i;
             }
-            mDayNumbers[i] = NumberFormat.getInstance().format(time.monthDay++);
-            time.normalize(true);
+            time.setDay(time.getDay() + 1);
```
Gitsaibot, line 301: Haven't had time to look at the new API but wouldn't `time.setDay(time.getDay() + 1);` have to come after `mDayNumbers[i] = NumberFormat.getInstance().format(time.getDay());` for it to behave the same as time.monthDay++ ?

chirayudesai: Yes, you're right!

Fixed.

chirayudesai:
> Edit: I think this comes from your changes in src/com/android/calendar/event/CreateEventDialogFragment.java

AOSP removed custom options from `format()`

It could potentially be re-added using `SimpleDateFormat` for this string.

mikeNG:
> 

This should be fixed by https://github.com/Etar-Group/Etar-Calendar/pull/1047/commits/12f41043c4a7062c7a1e5f5211b1c6a4a71569c7

Screenshot_20211221-201817_Calendar

redfast00:
Does this fix the year 2038 issue (https://github.com/Etar-Group/Etar-Calendar/issues/353)?

jspricke:
This is very welcome, thanks for working on it. We just had some releases to fix Etar for Android 12, once that is more stable we can continue here.

chirayudesai:
We've had a couple people review and test this downstream, 319630: Import internal Time API class | https://review.lineageos.org/c/LineageOS/android_packages_apps_Etar/+/319630

chirayudesai:
APK: https://github.com/chirayudesai/Etar-Calendar/actions/runs/1974757990

Gitsaibot (approved):

PR #1060, opened 2021-12-20
Sync general_preferences.xml

jspricke:


jspricke:
This makes -v26 and -v29 equal, I gues we can delete one but I did not find which one.

Gitsaibot (approved):

Issue #1069, opened 2022-01-03
Share Event-Entry with MIME="text/calendar" instead of MIME="application/octet-stream"


k3b:
I am currenly implementing a android-send-CalendarEntry plugin [CalEF (Calendar Entry Formatter)](https://github.com/k3b/calef)

that translates the human-unreadable technical format "ICS" (or "VCS") to human readable plain text.

When testing CalEF i realized that Etar-1.0.27 shares calendar events in format "ics-file" with MIME="application/octet-stream".

Is it possible to share an Event-Entry with MIME="text/calendar" (instead of MIME="application/octet-stream") ?

That would allow my app (and other apps) to selectivly receive "SEND CalendarEntries" ?

Note I also testet with other apps:

* [SimpleCalendar](https://github.com/SimpleMobileTools/Simple-Calendar) also uses -ics format with MIME="text/calendar" 
* My build-in stock calendar (LG-Phone with android-10) uses vcs files with MIME="text/x-vcalendar"

-------

How to reproduce:

* Make shure that [intentintercept from FDrdoid](https://f-droid.org/en/packages/de.k3b.android.intentintercept) Version 4.0 or newer is installed.
* Open Etar
* Click on a Calendar Event
* => The Calendar Event Details Page opens
* In the Toolbar of etar Click on the Share button
* => The Share Calendar-Event-Destination page opens
* Select "intentintercept" as share destination
* "intentintercept" opens with the share details.

PR #1070, opened 2022-01-03
Fix MIME type of shared event (Closes: #1069)

jspricke:


src/com/android/calendar/EventInfoFragment.java
```diff
@@ -1228,7 +1228,7 @@ private void shareEvent(ShareType type) {
                         mmsShareIntent.setAction(Intent.ACTION_SEND);
                         mmsShareIntent.setPackage("com.android.mms");
                         mmsShareIntent.putExtra(Intent.EXTRA_STREAM, vcsFile);
-                        mmsShareIntent.setType("text/x-vcalendar");
```
Gitsaibot, line 1231: Are you sure about this? // The MMS app only responds to "text/x-vcalendar" so we create a chooser intent...

jspricke: not sure but I'm not able to share with the MMS app either way. Does it work for you?

Gitsaibot: I don't use mms so I can't test. I only found this: https://stackoverflow.com/questions/14746872/what-is-the-correct-mime-type-for-vcalendar

jspricke: ok, let's leave this as is then.

Gitsaibot (approved):

PR #1078, opened 2022-01-08
switch widgetdesign to android 12 reference

newhinton:
This changes the widget-design to one that closely resembles the android 12 reference widgets.

grafik

Todo:

- [ ] Make this backwards compatible
- [ ] fix paddings
- [ ] use more color
- [ ] clean up service


newhinton:
It seems that the onClick event for widget_items are not opening the event. Is this a known bug with the widget?

Issue #803, opened 2020-11-04
Event colors are darker than what they should (calendar color)


llevrel:
I expect the calendar apps to use the color of each calendar to display the events. This is the case in e.g. Google Agenda and Simple Calendar. Etar on the contrary uses a different color. You can see the right color (defined e.g. on the CalDAV server) in the spot next to the calendar names, in the Settings page (of the left menu). The color used for the corresponding events is darker (maybe the hue is not changed).

For example, I have a calendar with bright yellow (#FFFF00). The events of this calendar show up in a kind of khaki. This makes the calendar views quite ugly. **I** am responsible for the choice of color, not Etar!

Gitsaibot:
>  I am responsible for the choice of color, not Etar!

Well then you probably have a solution how to recognize the white text font on the bright yellow background. For this reason the colors are darkened as far as I know.

llevrel:
> > I am responsible for the choice of color, not Etar!
> 
> Well then you probably have a solution how to recognize the white text font on the bright yellow background. For this reason the colors are darkened as far as I know.

Oh, I understand. The design option of a mandatory white color for the text is quite uncommon... Couldn't there be a setting to choose between white or black text ? There is already a choice for light/dark/black themes.

PR #1091, opened 2022-01-20
Use real calendar colors with adaptive text color (dark/light)

sof4jazz:
This adds the option to use the real calendar/event colors without the dark overlay. To make it work with light background colors it detects the brightness and switches to black text color.

Works on all views including the widget.

Solves this issue: #803 

Feel free to test it. Any feedback very much appreciated.

Don't know if it's better to remove the toggle switch entirely and make this the default behavior, but existing users might wanna stick to the classic rendering...

Before / After
Screenshot_1642686103-small
Screenshot_1642686145-small

Screenshot_1642686130-small
Screenshot_1642686150-small

Screenshot_1642687242-small

### DIFF
--- a/src/com/android/calendar/DayView.java
+++ b/src/com/android/calendar/DayView.java
@@ -2861,7 +2861,7 @@ public class DayView extends View implements View.OnCreateContextMenuListener,
                 case Attendees.ATTENDEE_STATUS_ACCEPTED:
                 case Attendees.ATTENDEE_STATUS_TENTATIVE:
                 default:
-                    paint.setColor(Utils.getAdaptiveTextColor(this.getContext(), mEventTextColor, event.color));
+                    paint.setColor(Utils.getAdaptiveTextColor(mContext, mEventTextColor, event.color));
                     break;
             }
 
--- a/src/com/android/calendar/DayView.java
+++ b/src/com/android/calendar/DayView.java
@@ -2856,14 +2856,12 @@ public class DayView extends View implements View.OnCreateContextMenuListener,
                     paint.setColor(event.color);
                     break;
                 case Attendees.ATTENDEE_STATUS_DECLINED:
-                    paint.setColor(mEventTextColor);
                     paint.setAlpha(Utils.DECLINED_EVENT_TEXT_ALPHA);
-                    break;
                 case Attendees.ATTENDEE_STATUS_NONE: // Your own events
                 case Attendees.ATTENDEE_STATUS_ACCEPTED:
                 case Attendees.ATTENDEE_STATUS_TENTATIVE:
                 default:
-                    paint.setColor(mEventTextColor);
+                    paint.setColor(Utils.getAdaptiveTextColor(this.getContext(), mEventTextColor, event.color));
                     break;
             }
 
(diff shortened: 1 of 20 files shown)


src/com/android/calendar/DayView.java
```diff
@@ -2856,14 +2856,12 @@ private StaticLayout getEventLayout(StaticLayout[] layouts, int i, Event event,
...
                 case Attendees.ATTENDEE_STATUS_DECLINED:
-                    paint.setColor(mEventTextColor);
                     paint.setAlpha(Utils.DECLINED_EVENT_TEXT_ALPHA);
-                    break;
                 case Attendees.ATTENDEE_STATUS_NONE: // Your own events
                 case Attendees.ATTENDEE_STATUS_ACCEPTED:
                 case Attendees.ATTENDEE_STATUS_TENTATIVE:
                 default:
-                    paint.setColor(mEventTextColor);
+                    paint.setColor(Utils.getAdaptiveTextColor(this.getContext(), mEventTextColor, event.color));
```
Gitsaibot, line 2864: Why not mContext?

src/com/android/calendar/EventInfoFragment.java
```diff
@@ -2246,7 +2255,7 @@ private void setDialogSize(Resources r) {
     public void onColorSelected(int color) {
         mCurrentColor = color;
         mCurrentColorKey = mDisplayColorKeyMap.get(color);
-        mHeadlines.setBackgroundColor(color);
+        updateEvent(mView);
```
Gitsaibot, line 2258: I wonder if there is a better way to adjust the colors than reloading the whole view and if the saved resources justify the additional code? The more I think about it, the more I think that the way it is here is the best way.

sof4jazz: Yeah, I first updated the colors without calling updateEvent, but that didn't work for the timezone which is a inside span. My first idea was to extract that part (from line 1501) to a separate method and call that from both places. Might be the cleaner way than do the whole updateEvent...

src/com/android/calendar/widget/CalendarAppWidgetService.java
```diff
@@ -291,7 +291,7 @@ public RemoteViews getViewAt(int position) {
                 } else {
                     views = new RemoteViews(mContext.getPackageName(), R.layout.widget_item);
                 }
-                int displayColor = Utils.getDisplayColorFromColor(eventInfo.color);
+                int displayColor = Utils.getDisplayColorFromColor(mContext, eventInfo.color);
```
Gitsaibot, line 294: int adaptiveColor = Utils.getAdaptiveTextColor(mContext, mStandardColor, displayColor);

src/com/android/calendar/widget/CalendarAppWidgetService.java
```diff
@@ -348,9 +348,9 @@ public RemoteViews getViewAt(int position) {
                     } else {
                         views.setInt(R.id.agenda_item_color, "setImageResource",
                                 R.drawable.widget_chip_responded_bg);
-                        views.setInt(R.id.title, "setTextColor", mStandardColor);
-                        views.setInt(R.id.when, "setTextColor", mStandardColor);
-                        views.setInt(R.id.where, "setTextColor", mStandardColor);
+                        views.setInt(R.id.title, "setTextColor", Utils.getAdaptiveTextColor(mContext, mStandardColor, displayColor));
```
Gitsaibot, line 351: views.setInt(R.id.title, "setTextColor", adaptiveColor);

sof4jazz:
Thanks, I'll have a look

sof4jazz:
@Gitsaibot What's the difference between those two? Top is all-day?
image

Gitsaibot (approved):

Issue #620, opened 2020-01-13
Reminder date on a whole day appointment - [Feature Request]


emgerber:
It would be great if I can make a whole day appointment and the reminder date is for example at 9 AM and not only before or at midnight.

woheller69:
In the Etar version of LineageOS this worked at least for whole day appointments imported from .ics

woheller69:
I created a pull request #1113 which should restore the display of reminders after start of the event.
So if you import an event, e.g. with https://github.com/SufficientlySecure/calendar-import-export where the reminder is
set at 9 AM this should work.
Of course it would be nicer if Etar could create such reminders with negative values.

Garbulix:
+1 for this

Etar lacks the option to pick a reminder on exact *datetime* for a whole day appointment. I can only chose a standard reminder like "10 minutes before", "2 hours before", "1 week before" etc. In such cases reference time is 00:00 so then I will receive a reminder e.g. 1 week before *at midnight*, when in most cases a person is sleeping.

This type of reminder selector is available in Google Calendar. I guess this is also a reason a datetime reminders set in the Google Calendar render poorly in Etar, like "327 hours" for 2 weeks before at 9:00 and "159 hours" for 1 week before at 9:00.

PR #1113, opened 2022-03-06
Repair display of reminders with negative values

woheller69:
E.g. for birthdays (whole day event) I have set a reminder at 9:00am, so after the start of the event.
The reminder works, but is not displayed currently because if
minutes < 0 reminders are ignored in EventViewUtils.java
(in the past that worked but it was changed with commit 032913f87c4256909d1911a9f1959849232d93a1)

### DIFF
--- a/src/com/android/calendar/event/EventViewUtils.java
+++ b/src/com/android/calendar/event/EventViewUtils.java
@@ -49,14 +49,7 @@ public class EventViewUtils {
         Resources resources = context.getResources();
         int value, resId;
 
-        if (minutes < 0) {
-            value = 0;
-            resId = R.string.no_reminder_label;
-
-            String format = resources.getString(resId, value);
-            return String.format(format, value);
-
-        } else if (minutes % 60 != 0 || minutes == 0) {
+        if (minutes % 60 != 0 || minutes == 0) {
             value = minutes;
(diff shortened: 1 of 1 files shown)


woheller69:
If you open an event with a reminder after start this time is displayed at the top, e.g. "- 9 hours", like it used to be in the past.
If you create a new event the drop down menu is as before my change.

woheller69:
Screenshot_20220306-193314_Etar

Screenshot_20220306-193401_Etar

woheller69:
Another question related to this: what about adding e.g. a checkbox like "after start of event" below the dropdown list and allow creation of such reminders as requested in #620?

woheller69:
... or add an entry "Manual" at the end of the dropdown list and when selected let the user enter the time for the reminder as desired via another dialog. If that is of interest I might try to realize it and come up with another pull request

woheller69:
> Looks good, thanks for the screenshots! Will test this today and tomorrow and then merge.

Are you still planning to merge this? #1117 is an improved version - in my view - which is easier to understand (no minus sign) and also allows creation of such reminders

woheller69:
> Not that, but #1117. Haven't had time to test it thoroughly yet. Maybe this weekend.

Excellent :-)

PR #1116, opened 2022-03-07
Set individual reminder, before or after start of event

woheller69:
As discussed in #1113 and requested in #620 here is a draft for a possibility to manually add a reminder in minutes before or after start of the event.
So far it only works for events that already exist. Due to the mechanism of reading the reminder times from the labels in the dropdown box it is a bit complicated to realize.

What I have done is add another entry "Manual" to the dropdown list. When selected an alert dialog is opened which allows to set the time in minutes and set before/after start via a checkbox. Then the reminders are saved and the fragment is closed. After opening it again it will show the added reminder.

Screenshot_20220307-131320_Etar

Maybe someone with more experience in the Etar source code can help improve and finalize it.

### DIFF
--- a/res/layout/manual_reminder.xml
+++ b/res/layout/manual_reminder.xml
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:orientation="vertical"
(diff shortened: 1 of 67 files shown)


woheller69:
I am closing this as the solution proposed in #1117 is much simpler and better in my view

Issue #1181, opened 2022-06-22
Notifications unreliable


leohoppergit:
Recently notifications have become super unreliable. In my experience, almost none of the events with notifications set, actually notify.

Garbulix:
I also have that problem on my LineageOS 19 and my friend (with LineageOS 18.1) also.
I commented mentioned Pull Request, because I believe its not the proper fix.

kuba-orlik:
The notifications are unreliable to the point of being useless. Here's an example:

![image](https://user-images.githubusercontent.com/2697916/202790841-33197c84-2c61-4ea3-9278-a4ae2f31f301.png)

The notification appeared.. a day later, at a totally unrelated hour....

Issue #1179, opened 2022-06-13
No notification at all, all of a sudden


Kalytis:
For a couple of weeks now, Etar basically refuses to give me any notification from my events on my phone. No idea what's the cause.

Battery notification is disabled.

Sorry for the lack of precision, but I have to admit, I'm puzzled.

Gitsaibot:
Maybe #1026 helps...

Gitsaibot:
Maybe #1026 helps...

Kalytis:
@Gitsaibot I'm not sure it's related to this. I've disabled battery optimization from day 1, to this day. It's still not helping.

Kalytis:
I've seen people suggest to add the Etar widget on homepage to fix the notification issue : can confirm it DOES NOT work for me either.

Gitsaibot:
I think that in your case only a debug log will help to identify the problem.

Kalytis:
Sure, could someone show me how to do this please ?

Kalytis:
Alright, absolutely not sure whether this will be helpful, but still.

`07-20 02:55:00.002  1627  1955 V AlarmManager: Triggering alarm #0: 3 when =92835464 package =android operation = null listenTag =TIME_TICK flags =0x1
07-20 02:55:00.003  1627  1955 V AlarmManager: Triggering alarm #1: 0 when =1658278500000 package =com.android.providers.calendar operation =*walarm*:android.intent.action.EVENT_REMINDER flags =0x5
07-20 02:55:00.003  1627  1955 V AlarmManager: Triggering alarm #2: 1 when =1658278500000 package =cz.martykan.forecastie operation =*alarm*:cz.martykan.forecastie.UPDATE_TIME flags =0x9
07-20 02:55:00.022  2612  2807 I ClockCtrl:  schedule next: 59978
07-20 02:55:00.022  2612  2807 I OpKeyguardUpdateMonitor: onTimeChanged
07-20 02:55:00.031  2612  2612 D KeyguardUpdateMonitor: received broadcast android.intent.action.TIME_TICK
07-20 02:55:00.041  2612  2612 D KeyguardUpdateMonitor: handleTimeUpdate
07-20 02:55:00.081  1627  2400 D SensorService: FusionLight, after outEvent->light = 0.000000[LB]
[...]

Issue #1163, opened 2022-05-16
No reminder notifications when app is not running


hockeymikey:
I've tried a few versions from 1.0.24 to 1.0.30 and it's the same result. If Etar is not running in the background, there aren't any notifications. So if the phone is sleeping, I don't get them. It is disabled in doze, so that isn't the issue. It used to work fine, but only recently started acting up, I do not know why. It's a huge issue not being able to get reminders for important events.

My device is a Sony Xperia xz1c running Android 8.1. Logcats don't show much unless I'm missing something.

markcellus:
Looks like this is a duplicate of #1026

Issue #1142, opened 2022-04-11
Crashes in background on Android 12


rugk:
After a recent update to the calendar from CalyxsOS (which is a fork of this one here) crashes now:

```
FATAL EXCEPTION: main
Process: org.calyxos.etar, PID: 5160
java.lang.RuntimeException: Unable to start receiver com.android.calendar.alerts.AlertReceiver: android.app.ForegroundServiceStartNotAllowedException: startForegroundService() not allowed due to mAllowStartForeground false: service org.calyxos.etar/com.android.calendar.alerts.AlertService
	at android.app.ActivityThread.handleReceiver(ActivityThread.java:4384)
	at android.app.ActivityThread.access$1600(ActivityThread.java:256)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2102)
	at android.os.Handler.dispatchMessage(Handler.java:106)
	at android.os.Looper.loopOnce(Looper.java:201)
	at android.os.Looper.loop(Looper.java:288)
	at android.app.ActivityThread.main(ActivityThread.java:7870)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1003)
Caused by: android.app.ForegroundServiceStartNotAllowedException: startForegroundService() not allowed due to mAllowStartForeground false: service org.calyxos.etar/com.android.calendar.alerts.AlertService
	at android.app.ForegroundServiceStartNotAllowedException$1.createFromParcel(ForegroundServiceStartNotAllowedException.java:54)
	at android.app.ForegroundServiceStartNotAllowedException$1.createFromParcel(ForegroundServiceStartNotAllowedException.java:50)
	at android.os.Parcel.readParcelable(Parcel.java:3333)
	at android.os.Parcel.createExceptionOrNull(Parcel.java:2420)
	at android.os.Parcel.createException(Parcel.java:2409)
	at android.os.Parcel.readException(Parcel.java:2392)
	at android.os.Parcel.readException(Parcel.java:2334)
	at android.app.IActivityManager$Stub$Proxy.startService(IActivityManager.java:6028)
	at android.app.ContextImpl.startServiceCommon(ContextImpl.java:1855)
	at android.app.ContextImpl.startForegroundService(ContextImpl.java:1831)
[...]

PR #1184, opened 2022-06-28
Etar: Request user to disable battery optimizations

theimpulson:
## Summary

This MR:

- Prompts users to disable battery optimizations aka doze for Etar to allow it to operate freely in the background,
- Guards the service to ensure it only runs if optimizations are disabled

Disabling battery optimizations makes sense considering we are seeing numerous issues even after using exact alarms complaining about notifications delay or missing completely. This also allows Etar to work properly on Android 12 due to new foreground service restrictions.

## Issue(s)

- #1142,
- #1181,
- #1179,
- #1026,
- #1163,
- #789

## Test

Install build with this MR, ensure that:
- New installs prompt the user to disable optimizations once permissions have been granted,
- Existing installs show an info icon on the toolbar which prompts the user to disable the optimizations.

Ensure Etar no longer crashes on Android 12 and can send event notifications if battery optimizations are disabled.

### DIFF
--- a/res/drawable/ic_info_outline.xml
+++ b/res/drawable/ic_info_outline.xml
@@ -1,5 +1,26 @@
-<vector android:height="24dp" android:tint="?attr/iconSettingsAbout"
-    android:viewportHeight="24.0" android:viewportWidth="24.0"
-    android:width="24dp" xmlns:android="http://schemas.android.com/apk/res/android">
-    <path android:fillColor="#FF000000" android:pathData="M11,17h2v-6h-2v6zM12,2C6.48,2 2,6.48 2,12s4.48,10 10,10 10,-4.48 10,-10S17.52,2 12,2zM12,20c-4.41,0 -8,-3.59 -8,-8s3.59,-8 8,-8 8,3.59 8,8 -3.59,8 -8,8zM11,9h2L13,7h-2v2z"/>
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (C) 2022 The Calyx Institute
+
+     Licensed under the Apache License, Version 2.0 (the "License");
+     you may not use this file except in compliance with the License.
+     You may obtain a copy of the License at
+
+          http://www.apache.org/licenses/LICENSE-2.0
+
+     Unless required by applicable law or agreed to in writing, software
+     distributed under the License is distributed on an "AS IS" BASIS,
+     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+     See the License for the specific language governing permissions and
+     limitations under the License.
+-->
+
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:tint="?attr/colorControlNormal"
+    android:viewportWidth="24.0"
+    android:viewportHeight="24.0">
+    <path
+        android:fillColor="@android:color/white"
+        android:pathData="M11,17h2v-6h-2v6zM12,2C6.48,2 2,6.48 2,12s4.48,10 10,10 10,-4.48 10,-10S17.52,2 12,2zM12,20c-4.41,0 -8,-3.59 -8,-8s3.59,-8 8,-8 8,3.59 8,8 -3.59,8 -8,8zM11,9h2L13,7h-2v2z" />
 </vector>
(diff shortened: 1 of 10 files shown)


theimpulson:
@Gitsaibot please review when possible

res/drawable/ic_info_outline.xml
```diff
@@ -1,5 +1,26 @@
...
+
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:tint="?attr/colorControlNormal"
+    android:viewportWidth="24.0"
+    android:viewportHeight="24.0">
+    <path
+        android:fillColor="@android:color/white"
+        android:pathData="M11,17h2v-6h-2v6zM12,2C6.48,2 2,6.48 2,12s4.48,10 10,10 10,-4.48 10,-10S17.52,2 12,2zM12,20c-4.41,0 -8,-3.59 -8,-8s3.59,-8 8,-8 8,3.59 8,8 -3.59,8 -8,8zM11,9h2L13,7h-2v2z" />
```
Gitsaibot, line 25: Why are changes needed here?

Gitsaibot, line 25: Why are changes needed here?

theimpulson: The major change is adding tint parameter to play nice with theme colors. Later on, the file has been reformatted using Android Studio to improve readability.

theimpulson: Seems duplicate comment from above, pasting same response:

The major change is adding tint parameter to play nice with theme colors. Later on, the file has been reformatted using Android Studio to improve readability.

theimpulson: Fixed. Added the tint used by the "About Info" screen to the image widget.

theimpulson: The drawable file is still used. What I meant is that I added the color in the layout's image widget. This file is still used.

Gitsaibot: I mean this reference "?attr/iconSettingsAbout"

theimpulson:
> I have here a Oneplus 5 with Android 10 and there the info icon is not displayed even if I have activated the battery mode Optimize or Smart Control. I think your commit doesn't work with manufacturer specific power saving settings. There's probably not much you can do about that.
> 
> When I change the power saving settings on android 12, the icon appears in the menu bar anyway. Have you tested what happens when the automatic battery saving mode is activated when the battery level drops below 10 or 5 percent?

As you say there is not much we can do about manufacturer-specific power-saving settings. I tested the application on an emulator running Android 12 OS. Click on the icon, it will ask you for permission and when granted it hides it.

res/drawable/ic_info_outline.xml
theimpulson: Ah, you are right. I seem to have to use another attribute instead of the one you pointed out. I switched to that one now. It should be fine now.

Gitsaibot (approved):

rugk:
Ref for anyone coming here and having other/manufacturer specific issues/optimisations or so. There is a page called https://dontkillmyapp.com/?app=Etar for you that can guide you to disable those, too.

Garbulix:
@theimpulson 
are you sure requesting disabling battery optimization is needed. Maybe there can be some altering in code performed, instead of asking for such action?
I doubt it, because of these reasons:
1. I am using Loop habits tracker (https://github.com/iSoron/uhabits) and its notifications are still reliable, even after updating to Android 12 (LineageOS 19) in my Xiaomi Mi A1. I didn't change battery optimization settings for this app (it has default "optimized" setting. Please note here Loop isn't Lineage's system app.
2. Etar is Lineage's system app, so I *guess* in that case it should have some special privileges that make the app doesn't need changing battery settings. 
3. My friend uses LineageOS 18.1 based on Android **11** (which, I assume, doesn't have these special backgroud app restrictions that you mentioned) and he also have problems with notifications reliability, which haven't occured earlier.

theimpulson:
> 

@Garbulix, yes, it is needed as Etar uses foreground service to work with notifications, whereas the repository you linked doesn't seem to use foreground services. As for being a system app, you don't need to grant it manually in LineageOS, as it is granted by default.

Garbulix:
So how is that possible mentioned Loop (which shows me reminders at given hours on given weekdays) app works without problems even on Android 12 and why Etar as system app on my Lineage works unreliable?

Garbulix:
I reread what you wrote and I see I omitted the part about foreground service - this part I don't understand as I am not a programmer. Even if this part may be omitted, I still don't understand why Etar became unreliable on Android 11 and why it became unreliable even as Lineage's default calendar app

PR #1187, opened 2022-07-04
Refactor and improve Gradle configuration

theimpulson:
## Summary

- Sync with the latest templates,
- Migrate to plugins & lint block globally,
- Switch to new dependency and plugins management,
- Split external dependency configurations into common and apply it,
- Switch to new Sdk version properties,
- Bump Gradle version to the latest,
- Enable Java 11 lang features.

## Test

- Sync and build using Gradle in Android Studio. Notice build completes without any issues and generated artifact works fine.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,199 +1,48 @@
-buildscript {
-    ext.kotlin_version = '1.6.10'
-    ext.kotlin_coroutines_version = '1.6.0'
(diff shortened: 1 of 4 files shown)


luk1337:
We don't use Gradle on lineage so this change won't really change anything.

Issue #749, opened 2020-07-27
Feature Request: Copy & Move Events in same or different calendar


e-t-l:
Related to #666, it would be great to be able to duplicate an event (or even multiple events at a time!) as well as move events from one calendar to another.

petrhejna:
This also provides workaround for "edit calendar". One can duplicate event into different one and then delete original one.

dilyanpalauzov:
For events with attendees managed over CalDAV (“scheduled events”) either the move must be done with the HTTP MOVE verb (to my knowledge currently not supported by DavX5), or all attendees will get a cancellation notification , sent by the server, for the old calendar, and a new meeting invitation for the new calendar (this is what the DELETE and PUT http verbs request)

To avoid the above confusion, Cyrus Imap disallows to have for the same users two events in different calendars with the same UID.  So if this is implemented for events without attendees, and the order of DELETE and PUT is unclear, the UID must be changed.

SilverPuppy:
I second the need for this feature. My current workaround is just to reschedule the old event, but that destroys the record of the old event. Not ideal. Please implement this as you are able.

MathieuSchopfer:
I think #87 should be implemented first. It would provide the base for this task.

Cwpute:
Regarding the asked feature of moving events between calendars, this has éeen requested before in #62 and is being worked on (see draft in linked issue).

Thatoo:
Can't this issue be closed now?

Issue #666, opened 2020-03-03
Copy a task? 


bjlockie:
Is there a way to copy an entry?

just-Nob:
I like to have that possibility as well.


I'm hardly missing the possibility to copy/duplicate a single date entry.
This should work similar to moving an entry, but the original entry still stays unchanged.
This is a feature I often like to use for a series of similar dates/appointments (like a training course), but with slightly different content, or where the date's day and/or time doesn't follow any regularity (example: first date of sequence is on Monday 8:00 a.m for 3h., second on Wednesday 10:00 a.m for 1,5h., third on Thursday 4:30 p.m. for 1h, fourth on Friday 9:15 a.m. for 2h etc.).

Hubert21:
Hi
I second that request!

Knote9:
Absolutely Agree.
It's the only gap I have with the app.
Please prioritise! Thanks.

SilverPuppy:
I commented in #749, but will comment here too. This is important to me as well. Not having it requires rescheduling a previous event, destroying its record. I would love to have this available.

tkrypton:
Hi, yes, this feature would make dump the google calendar completely!

PR #1248, opened 2022-10-18
Event duplication 

### LINKED ISSUE #87: Copy Events
Would be nice if one could copy events to a certain date.

jonas-haeusler:
testing and feedback is very welcome :-)

resolves #87, #666, #749.

the duplication option also provides a better workaround to move events between calendars.

i modified `event_info_title_bar.xml` a bit to make the top bar less cluttered by hiding items if android thinks there is not enough room for them, and by reordering the items by (my subjective) priority. let me know if this is not wanted.


jonas-haeusler:
> With your change only 2 icons are displayed although there would be room for more.

this is done deliberateley to follow material design guidelines, see https://m3.material.io/components/top-app-bar/guidelines#224e8ca0-4177-45f4-9f4f-eef1fbb0e4b2.

in addition to that, when using `app:showAsAction="always"` the top app bar will not be displayed correctly on smaller screen sizes:
grafik

which is also stated by the android docs: https://developer.android.com/guide/topics/resources/menu-resource
> ***android:showAsAction***
> `always`
> Always place this item in the app bar. Avoid using this unless it's critical that the item always appear in the action bar. Setting multiple items to always appear as action items can result in them overlapping with other UI in the app bar.

jspricke:
I would order them by priority and use `ifRoom`. My order would be:
- edit
- delete
- duplicate
- share
- export
- change_color

Gitsaibot (approved):

PR #1259, opened 2022-11-17
Day view should always reload on day change

### LINKED ISSUE #1194: Event Display Lag in Day View When Scrolling a Few Day Ahead or Behind
I have a tremendous display lag when I scroll in Day View a few days ahead or behind. I see an empty day when I know there are events there and they won't appear until I back out of the app then come right back in and then they refresh. If I wait, sometimes it takes 20 seconds to display the events on a day but sometimes it never does and I still have to back out of the app and come back in. The refresh option in the menu does nothing.

I deleted the data for the app and that fixed it for a few days then it became massively lagged again.

This is really annoying. I tried enabling contacts permission but that didn't work. I wonder why the refresh on new days a few days later or earlier is sooooo slow. Any one else having this problem?

This might be related to issue #434

I am using version 1.0.30 on LineageOS 19.1. It was happening on a Pixel 3 and now a Pixel 5 too.

t-m-w:
Day view was avoiding reloading when swiping between days because it based its reload decision on the starting day of the week.

Resolves #1194

### DIFF
--- a/app/src/main/java/com/android/calendar/DayView.java
+++ b/app/src/main/java/com/android/calendar/DayView.java
@@ -2019,18 +2019,13 @@ public class DayView extends View implements View.OnCreateContextMenuListener,
         mPrevSelectedEvent = null;
         mSelectedEvents.clear();
 
-        // The start date is the beginning of the week at 12am
-        Time weekStart = new Time(Utils.getTimeZone(mContext, mTZUpdater));
-        weekStart.set(mBaseDate);
-        // Avoid zero when Sunday is selected as the start day of the week.
-        if (mFirstDayOfWeek == 0) {
-            mFirstDayOfWeek = 7;
-        }
-        weekStart.setWeekDay(mFirstDayOfWeek);
-        weekStart.setHour(0);
-        weekStart.setMinute(0);
-        weekStart.setSecond(0);
-        long millis = weekStart.normalize();
+        // The start time is the beginning of the day at 12am
+        Time dayStart = new Time(Utils.getTimeZone(mContext, mTZUpdater));
+        dayStart.set(mBaseDate);
+        dayStart.setHour(0);
+        dayStart.setMinute(0);
+        dayStart.setSecond(0);
+        long millis = dayStart.normalize();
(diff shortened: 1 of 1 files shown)


t-m-w (commented):
[Edit: Updated patch to change to dayStart.] Hmm. With this change, it looks like weekStart is not used anymore, so that all could be removed. I don't understand what it was ever trying to do, unless it was meant to accommodate Week view with no regard for Day view.

landry314:
i would love to but i cant find an apk at that link.

landry314:
im using octodroid on my android and i cant see how to download that. i can try my mac.

landry314:
ok, i got the apk on my phone and tested it and it worked! i even compared the debug app to the app i have installed and noticed failed refreshes on the normal version while the debug version refreshed fine on the exact same days.

FIXED

landry314 (approved):
ok, i got the apk on my phone and tested it and it worked! i even compared the debug app to the app i have installed and noticed failed refreshes on the normal version while the debug version refreshed fine on the exact same days.

FIXED

Gitsaibot (approved):

landry314:
ok, hold on. i am getting refresh bugs in the debug version. i've been on vacation and not using it but now that i am, it isn't good. now when the day doesn't load, simply backing out of the app doesn't fix it. it takes about 5 seconds to refresh. this is happening more than before, maybe now that the days are filled with events it is occuring. the event blocks on the day view are flashing when you scroll quickly back and forth but when you stop nothing is displayed and never comes until you leave the app for 5 seconds. :-( . this is worse. sorry i said it was fixed above. that was a preliminary test. now i am really using it and it does not work.

landry314:
sorry, as i explained in the issue that led to this commit, this refresh error is only happening when "Use home time zone" is on. Otherwise the commit DID fix the problem. I just compared the debug version with the latest released apk 1.0.30 when both of the apps had "Use home time zone" OFF and the 1.0.30 version DOES have this bug while the debug version does NOT. therefore this commit should stand and does solve the problem.

PR #1283, opened 2022-12-24
Add aarGen dependency to app build

jspricke:


jspricke:
Needed to fix F-Droid builds:
https://monitor.f-droid.org/builds/log/ws.xsoh.etar/33#site-footer

Gitsaibot (approved):

Issue #1360, opened 2023-03-15
No alerts


gerdmueller1:
Any idea, since some time I do not get any Etar pop up for alerts. Schedulea run through without me getting an alert pop up. Thank you

bjlockie:
Neither do I.
I kept forgetting to come here.
Android 11.

gerdmueller1:
Thank you. The second one behind the link of the other post works for me. Great!

woheller69:
If that works, also the latest release should work for you

CS1o:
> Try the solutions in this issue.

The only Solution would be a workaround apk provided from woheller69 linked here but they are no longer available.
So there is no active Solution to this problem :/

woheller69:
Does V1.0.33 from F-Droid work?

CS1o:
> Does V1.0.33 from F-Droid work?

Seems like its working again with 1.0.33, thx.
But would love a fix for the new Version.

woheller69:
Does the apk from here work?

https://github.com/Etar-Group/Etar-Calendar/suites/13126535761/artifacts/713157153

CS1o:
> Does the apk from here work? It is built via reopened #1363
> 
> https://github.com/Etar-Group/Etar-Calendar/suites/13126535761/artifacts/713157153

Hey, i tested the apk, and it works! After that i uninstalled it and reinstalled the normal 1.0.34 again to compare.
Strangely it works now too with Notifications?!
idk if its because i granted your app needed more perms and it used the same for the other version then too.

Kalytis:
For me, it used to work for a time whenever I'd install a new version, until it stopped again.

hathagat:
Same here! Notifications ar not working at all.

Etar 1.0.34
current Android 13
Pixel 7

The apk provided by @woheller69 did the trick! Thanks a lot!

jisse44:
Hello, i'm very surprised tsee new version 1.0.35 without fix for a 9 months identified bug.
rollback to @woheller69 build ...

lammel-hub:
Same here. It used to work, but with the android version that introduced a tick box option for notifications after the event has started (1.0.35?), I no longer get ANY notifications from Etar. 

Sadly, if this is not remedied soon I will have to switch app, as this is a fundamental functionality for me.

I am a regular user who do not understand the references to APK:s above. Yes, I have donated to the project.

jspricke:
> Yes, I have donated to the project.

[...]

Issue #1317, opened 2023-01-09
BUG: Functionless alarms and postponement (no sound, no reminders) 


Moostek:
**With the new update 1.0.34 the alarms DO NOT WORK COMPLETELY: no sounds, nothing.** 

Is there anybody who care and read the reported issues and bugs? 
Nobody did react on my (and other) reports. As a result, the application is completely **useless** now.. 
I really do not understand this precedure "the new update -> nothing works". 

p.s.: maybe it works on developer's phone, but i face these bugs, and with the new update, the reminders stopped to work completely. The previous version 1.0.33 suffers with [the postponement bug](https://github.com/Etar-Group/Etar-Calendar/issues/1307) of the events with no sound and visual reminder (occasionaly it works = not reliable).

jspricke:
Yes we read the issues but we are all voluntary so please contribute if you have a problem. For me the app works on an Pixel 6 with CalyxOS 4.5.0. Also note due to the many different versions of Android there could be a lot of causes not related to Etar

Gitsaibot:
I think there is a fundamental misunderstanding here. Etar is open source and not a commercial product. Open source is based on collaboration, but it seems that everyone here wants to be a beneficiary and not contribute anything. 

When I have time I read the bug reports here. But what should I do with one like this? Apart from complaints, it doesn't contain any relevant info for me. No debug log. No information about the hardware used. Just nothing but complaints.

Moostek:
[...]

PR #1363, opened 2023-03-23
Use permission SCHEDULE_EXACT_ALARM instead of USE_EXACT_ALARM

woheller69:
Revert https://github.com/Etar-Group/Etar-Calendar/commit/9555bdba889a40df892fe47f51aac6b72aadeb0e to fix issues with notification e.g. #1317, #1360, etc


jspricke:
I disagree here as it makes it work with new Android versions and is the recommended way according to Google. I would propose to do more research if this is really the problem (like how is it implemented, which library, are there other users,..) and report to Google in case.

woheller69:
SCHEDULE_EXACT_ALARM also works with new Android versions, so why kill all older Androids?

jspricke:
it never worked reliably.

woheller69:
but wasn't that due to battery optimization which has been fixed?

woheller69:
> @woheller69 What does the change have to do with the older devices? Up to api 32 SCHEDULE_EXACT_ALARM is still used even with the change. Only the devices from api 33 use USE_EXACT_ALARM

Obviously it does not. I asked the users to test the apk from CI before and after this change and it does not work after the change anymore.

jspricke:
> > @woheller69 What does the change have to do with the older devices? Up to api 32 SCHEDULE_EXACT_ALARM is still used even with the change. Only the devices from api 33 use USE_EXACT_ALARM
> 
> Obviously it does not. I asked the users to test the apk from CI before and after this change and it does not work after the change anymore.

Sounds like a good start to investigate why that logic does not work.

woheller69:
[SCHEDULE_EXACT_ALARM](https://developer.android.com/reference/android/Manifest.permission#SCHEDULE_EXACT_ALARM), the permission introduced in Android 12 for apps to schedule exact alarms, is no longer being pre-granted to most newly installed apps targeting Android 13 and higher (will be set to denied by default).
(From Android 14 changes)

So we would need to check :

Apps should call [AlarmManager.canScheduleExactAlarms()](https://developer.android.com/reference/android/app/AlarmManager#canScheduleExactAlarms()) to confirm that it has the appropriate permission.
If the app doesn't have the permission, invoke an intent that includes the [ACTION_REQUEST_SCHEDULE_EXACT_ALARM](https://developer.android.com/reference/android/provider/Settings#ACTION_REQUEST_SCHEDULE_EXACT_ALARM), along with the app's package name, to ask users to grant the permission.

woheller69:
For Android12+ add

    AlarmManager alarmManager = (AlarmManager) getSystemService(Context.ALARM_SERVICE)
    if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.S) {
            if (!alarmManager.canScheduleExactAlarms()) {
                Intent intent = new Intent();
                intent.setAction(Settings.ACTION_REQUEST_SCHEDULE_EXACT_ALARM);
                startActivity(intent);
            }
        }

bjlockie:
I have Android 11.

jspricke:
@woheller69 I think adding the intent code makes a lot of sense for older devices, can you open a pull request with it? Also having it tested on those devices would be great.

woheller69:
old devices do not need that. This is for Android 12+ devices to work with SCHEULE_EXACT_ALARM instead of USE_ECAXT_ALARM

woheller69:
and Android 12 normally does not need that either. Unless the user actively revokes that permission

CS1o:
> Oh, right Android 12 ~ older devices ;).

The fix would be needed for Android 13 and maybe higher too.
Please consider a merge :)

woheller69:
I added that part to my patch-2 branch:

AlarmManager alarmManager = (AlarmManager) getSystemService(Context.ALARM_SERVICE);
if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.S) {
        if (!alarmManager.canScheduleExactAlarms()) {
            Intent intent = new Intent();
            intent.setAction(Settings.ACTION_REQUEST_SCHEDULE_EXACT_ALARM);
            startActivity(intent);
        }
    }

Cannot compile it at the moment, as my Android Studio version is too old but I don't want to change it...
But I am sure it will work.

woheller69:
Just updated my Android Studio and tried to build it. But it fails. Probably some Gradle bullshit.

Namespace not specified. Please specify a namespace in the module's build.gradle file like so:

android {
    namespace 'com.example.namespace'
}

But a namespace is already defined.

Won't waste more time ...

Gitsaibot (approved):

PR #1370, opened 2023-04-09
UI Improvements

marreitin:
I generally improved the UI for the event editor and event info dialogs by aligning stuff better and making it look less overwhelming. Additionally added a simple fix for the info dialog crashing when a URL was added to the event by checking if the ExpandableTextView exists (it doesnt yet).


marreitin:
Hope one can see the changes, most of them are really just aligning stuff and replacing old pixel icons with vector assets so they don't appear blurry but rather scale nicely.
# New event editor
ScreenshotNew
# New timezone picker
ScreenshotNewTimezone
# Old event editor
ScreenshotOld
# Old timezone picker
ScreenshotOldTimezone

app/build.gradle
```diff
@@ -89,6 +89,7 @@ android {
     kotlinOptions {
         jvmTarget = "11"
     }
+    namespace 'ws.xsoh.etar'
```
Gitsaibot, line 92: what is the reason for the change?

app/src/main/res/layout/edit_event_all.xml
```diff
@@ -137,19 +137,17 @@
             android:id="@+id/change_color_new_event"
             android:layout_width="0dp"
             android:layout_height="wrap_content"
-            android:layout_marginEnd="32dp"
+            android:layout_marginEnd="24dp"
+            android:layout_marginTop="20dp"
             android:background="?android:attr/selectableItemBackground"
             android:contentDescription="@string/choose_event_color_label"
-            android:enabled="false"
+            android:enabled="true"
```
Gitsaibot, line 144: If I remember correctly there was a reason why it is off as standard. There are calendars where you cannot change the event color. With your change here, the icon will be displayed anyway.

app/src/main/res/layout/event_info_dialog.xml
```diff
@@ -13,10 +13,10 @@
      See the License for the specific language governing permissions and
      limitations under the License.
 -->
-<FrameLayout
-    xmlns:android="http://schemas.android.com/apk/res/android"
+<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
```
Gitsaibot, line 19: Why is the position changed here?

app/src/test/AndroidManifest.xml
```diff
@@ -15,8 +15,7 @@
 -->
 
 <!-- package name must be unique so suffix with "tests" so package loader doesn't ignore us -->
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="com.android.calendar.tests">
```
Gitsaibot, line 19: Why removing the package?

external/timezonepicker
Gitsaibot, line 1: Please commit your changes dirctly to our timezonepicker fork here: https://github.com/Etar-Group/standalone-calendar-timezonepicker

build.gradle
```diff
@@ -1,6 +1,6 @@
 plugins {
-    id 'com.android.application' version '7.2.1' apply false
-    id 'com.android.library' version '7.2.1' apply false
+    id 'com.android.application' version '7.4.2' apply false
```
Gitsaibot, line 2: I think the gradle changes have nothing to do with the actual pull request. Please make a separate pull request for that. Sometimes these changes lead to problems and then it is easier for us if this is separated.

app/src/main/res/drawable-hdpi/ic_menu_remove_field_holo_light.webp
Gitsaibot, line 1: Sure these drawables were not used anywhere else in the code?

marreitin: Yes, i searched the project and they should work everywhere

marreitin:
Maybe a general reply: Sorry for all the gradle changes, random packages removals etc, next time i'll watch out better that Android Studio doesn't randomly change stuff

build.gradle
marreitin: No just leave them out i fucked up setting it up in android studio... It then messed with all these configs, but those are not to be changed.

external/timezonepicker
marreitin: Will do

build.gradle
Gitsaibot: I can't just leave these out in Github. There is no function for that. You have to revert the changes for the corresponding files in your branch. Then upload again via "force push".

marreitin:
Cleaned up my commits, now there are no more files commited that shouldn't be changed.

marreitin:
Fixed, maybe i'll add an empty view inbetween, because on a very wide display there is the opposite issue where the drop-down-triangle (or whatever it's called) is miles away from the text.

app/src/main/java/com/android/calendar/EventInfoFragment.java
```diff
@@ -1648,7 +1648,7 @@ public boolean onTouch(View v, MotionEvent event) {
 
     private void updateExtended() {
         // Url
-        if (mEventUrl != null && !mEventUrl.isBlank()) {
+        if (mEventUrl != null && !mEventUrl.isBlank() && mUrl != null) {
```
Gitsaibot, line 1651: Can you tell me the steps to reproduce the crash.

marreitin:
Okay i probably missed this issue, just noticed this happening on my phone as well, seems like an issue with the spinners elements. I'll have a look at it as soon as possible.

app/src/main/java/com/android/calendar/EventInfoFragment.java
Gitsaibot: There is a url textview in event info layout. See https://github.com/Etar-Group/Etar-Calendar/commit/095a231e87a54ab8b9266cfa7c6386e20da46d4d

So i think we dont need this here.

marreitin: [Removed the change] Probably crashed because of some different issue or i just deleted it by accident whatever...

marreitin:
Found the issue with the spinner: The notification-type-spinners gravity is set to `top|end`, and when the text inside is large enough it only seems like it is aligned correctly but actually it is still aligned to the right. Now, what is better, aligning it to the right or to the left? I'd personally say left for consistency but maybe it is intentional design.

marreitin:
image
Apparently the error that caused the app to crash still persists, and I'll try to find at. As shown in the screenshot, somehow mUrl is null and not the url-view, and calling setText() on a null object causes the app to crash.

marreitin:
So i found the problem: On a phone the event_info.xml will be used for a full-screen event info, but on a tablet it uses the event_info_dialog.xml and that one didn't have a textview for the url which is why it was null. I copied over the textview from event_info.xml and now everything works (so you also see your added URLs on a tablet).

Gitsaibot (approved):

PR #1385, opened 2023-05-04
Prompt user to add calendar instead of account

theimpulson:
This ensures that an offline calendar is available as a choice to users. Otherwise users are confused and assume that Etar requires an online account to work.


Gitsaibot (approved):

theimpulson:
@Gitsaibot if everything is good, maybe we can merge this?

Issue #1319, opened 2023-01-13
Add controls, to jump to the next/previous month.


CodeCubeNeo:
Hello! I think that it would be a very useful feature to have buttons at the top (in the nav bar, smth like two arrow keys (<>)) for changing the month view. I think that scrooling isnt that accurate and is harder than just clicking a button (also scrooling should click the whole month in place, not just leave it to where you scrool). The funcionality is like that in the simple mobile tools's calendar app.

ArdakHere:
Hey! I have implemented something that you asked for. Take a look at it in my repository - https://github.com/ArdakHere/Etar-Calendar

l0nax:
@ArdakHere if you have implemented it, can you please create a PR with your changes?

CodeCubeNeo:
Yes please!

ArdakHere:
@l0nax @CodeCubeNeo hey! Check out the PRs 1390 and 1391.

Gitsaibot:
Please add the two files to one PR.

l0nax:
@ArdakHere is there a reason why you've closed those two PRs?

ArdakHere:
@l0nax no. It is just my first time submitting a PR. I was just trying to attach two files to a single PR.

CodeCubeNeo:
Any progress on releasing the new version with these changes?

CodeCubeNeo:
Hi. Any updates? @l0nax why is there no new version release after such a long time?

ArdakHere:
Hey! I am sorry, a bit busy now.

PR #1392, opened 2023-05-08
Issue #1319 solution

ArdakHere:
Added two functions for the two buttons < and >
Added two < > buttons to .xml file.


jspricke:
Did you test this?

Issue #1216, opened 2022-08-09
[BUG] date changes on edit


bjlockie:
Editing a recurring entry loses 1 day.

Create a monthly recurring entry called 'Test' (all day) for eg. Aug 8:
![image](https://user-images.githubusercontent.com/303873/183553956-56c22b00-4542-4a93-b0de-96b5eb3e0e23.png)

Edit the entry and it changes to Aug 7:
![image](https://user-images.githubusercontent.com/303873/183554321-5ece0201-f72b-4c84-862c-b6ac6b7e15b9.png)

Android 11
Etar 1.0.30
LG G7 One

glemco:
Possible duplicate of  #1173 and #1164, my rough guess is that the precise effect depends on your time zone

lmamane:
Likely fixed by GH-1405. Let me know.

Issue #1208, opened 2022-07-27
All-day events change date unexpectedly on edit.


rekweststory:
In Etar build 1.0.30 on an old android on eastern daylight time, create an all-day event (eg oct. 7), with no notification. Go back into the event to edit it to just add a location. Surprise! the date is now a day earlier (eg. oct. 6). If the date is reselected to the correct date, everthing is fine. (Offline calendar)

glemco:
Possible duplicate of  #1173 and #1164, my rough guess is that the precise effect depends on your time zone

lmamane:
Likely fixed by GH-1405. Let me know.

Issue #1173, opened 2022-06-01
Editing an all-day event makes it a day longer.


jtagcat:
1. Create an all-day event for any day, let's pick 2. june.
2. On calendar views, it is shown only on 2nd. While the text says 2nd-3rd.
3. Editing it, dates are 2nd-3rd.
4. Editing it again, dates are 2nd-4th.

While editing, toggling all-day changes day by one.

Parsnip:
I was just coming to report this. Every time you open an all day event for editing, the end date gets incremented by one. Even if you back out of the edit interface with the android back button, the new unintentionally incremented end date is saved.

This only happens with all day events as far as I've seen.

TOMats3:
Agree. I would add that on screen with event details / description, start date is put one day back.
To sum up:
Day / week / month / agenda screen: OK
Event details screen: start day -1 day
Event edit screen: end day +1 day

lmamane:
I remember I used to encounter this, but don't anymore. This might be because my system timezone (not app-specific home timezone) is UTC.

lmamane:
Could be fixed by GH-1405, not sure. Let me know.

Parsnip:
> Could be fixed by [GH-1405](https://github.com/Etar-Group/Etar-Calendar/pull/1405), not sure. Let me know.

I only just now noticed the update and it's fixed indeed, I at least can no longer reproduce any odd behavior with all day events. Great work! 👍❤️

IamJeffG:
> I only just now noticed the update and it's fixed indeed, I at least can no longer reproduce any odd behavior with all day events on my timezone

Same observation, this seems to be fixed for me right now too.  :+1:

Issue #1164, opened 2022-05-20
Event changes to the day before on edit


liperibeiro:
When I edit an event, I noticed the date is changed to the day before. I always have to remember to set the right date again before saving any change.
How to reproduce it:
1. Click on an event on the calendar
2. Click the pencil icon to edit it
At this point, in the Edit screen, the date is set to the day before. Let's say the event date was 05/25. When the edit screen opens it is set to 05/24.
I'm using v. 1.0.30 on Android 11

prg318:
I've also experienced this frequently with the latest version of etar and android 11

jspricke:
Can't reproduce on Android 12

aereaux:
I have a very reproducible, very similar issue, looks like #1173.  Is it possible it's a timzone issue or something?  Is there any debugging info I can provide?

glemco:
I can confirm it is a timezone issue, but the results are a little misleading (I didn't inspect the code much).
I tested with an all day event for the 18th of August with different time zones. Here are the results.
```
Created the event (all the same)
Begin 18th august - End 18th August
Viewing the event
Negative time zone: Begin 18th august - End 19th August
Positive time zone: Begin 17th august - End 18th August
Zero time zone: Begin 18th august - End 18th August
Editing the event
Negative time zone: Begin 18th august - End 19th August
Positive time zone: Begin 18th august - End 19th August
Zero time zone: Begin 18th august - End 18th August
```
As you can see the only case in which the event stays correct both viewing and editing is with a UTC time zone, the event looks as lasting 2 days if viewing in both plus and minus time zones, but with different offset. Editing the event causes the end to move for both in the same direction (saving without modifying makes the event last one more day).
Besides the untriggered modification mentioned above, in all cases changing the event from all-day to normal increases the end by a further day (e.g. 19th if UTC and 20th August otherwise)

Extended AOSP 12, etar 1.0.30 fdroid

Parsnip:
Is there any hope that this might get fixed since it seems the underlying issue is known? Etar seems to be one of the few if not the only calendar that has colored event support on non google accounts (e.g. through davx5) so it would be a shame to give that up.

glemco:
Unfortunately I'm not entirely sure the underlying issue is known. I'm running another android version (AICP 17.1, android 12.1) and it seems it's fixed for my time zone. Curiously enough if i put negative time zones, events are going backwards (still one day but the previous, in contrast with what I previously observed).
[...]

PR #1405, opened 2023-06-04
rework and robustify all-day events

### LINKED ISSUE #1398: creating new event from month view changes timezone to UTC
Steps to reproduce:
* long-click a day in month view to create a new event, immediately edit it
* change it to not be all day, notice how the timezone is shown as expected
* change the time of the event, notice how the timezone changed to UTC
This does not happen when editing an existing all-day event (though #382 does happen there). Unlike [various other issues](https://github.com/Etar-Group/Etar-Calendar/issues/1265), "use home time zone" is *off* in this issue.

Video of the issue: https://github.com/Etar-Group/Etar-Calendar/assets/101334480/e516dcc6-9b4a-447e-bb53-d2a55b6b616a

lmamane:
closes #1398 and bugs I haven't seen filed as open issues

### DIFF
--- a/app/src/main/java/com/android/calendar/Utils.java
+++ b/app/src/main/java/com/android/calendar/Utils.java
@@ -793,8 +793,9 @@ public class Utils {
         }
         recycle.setTimezone(Time.TIMEZONE_UTC);
         recycle.set(utcTime);
-        recycle.setTimezone(tz);
-        return recycle.normalize();
+        Time target = new Time(tz);
+        target.set(0, 0, 0, recycle.getDay(), recycle.getMonth(), recycle.getYear());
+        return target.normalize();
     }
 
     public static long convertAlldayLocalToUTC(Time recycle, long localTime, String tz) {
@@ -803,8 +804,9 @@ public class Utils {
         }
         recycle.setTimezone(tz);
         recycle.set(localTime);
-        recycle.setTimezone(Time.TIMEZONE_UTC);
-        return recycle.normalize();
+        Time target = new Time(Time.TIMEZONE_UTC);
+        target.set(0, 0, 0, recycle.getDay(), recycle.getMonth(), recycle.getYear());
+        return target.normalize();
     }
 
     /**
--- a/app/src/main/java/com/android/calendar/event/EditEventView.java
+++ b/app/src/main/java/com/android/calendar/event/EditEventView.java
@@ -658,19 +658,23 @@ public class EditEventView implements View.OnClickListener, DialogInterface.OnCa
         }
 
         if (mModel.mAllDay) {
-            // Reset start and end time, increment the monthDay by 1, and set
+            // Reset start and end time without touching date;
+            // in model, increment the monthDay by 1, and set
             // the timezone to UTC, as required for all-day events.
-            mTimezone = Time.TIMEZONE_UTC;
             mStartTime.setHour(0);
             mStartTime.setMinute(0);
             mStartTime.setSecond(0);
-            mStartTime.setTimezone(mTimezone);
-            mModel.mStart = mStartTime.normalize();
+            mStartTime.normalize();
+            Time modelStartTime = new Time(Time.TIMEZONE_UTC);
+            modelStartTime.set(0, 0, 0, mStartTime.getDay(), mStartTime.getMonth(), mStartTime.getYear());
+            mModel.mStart = modelStartTime.normalize();
 
             mEndTime.setHour(0);
             mEndTime.setMinute(0);
             mEndTime.setSecond(0);
-            mEndTime.setTimezone(mTimezone);
+            mEndTime.normalize();
+            Time modelEndTime = new Time(Time.TIMEZONE_UTC);
+            modelEndTime.set(0, 0, 0, mEndTime.getDay(), mEndTime.getMonth(), mEndTime.getYear());
             // When a user see the event duration as "X - Y" (e.g. Oct. 28 - Oct. 29), end time
             // should be Y + 1 (Oct.30).
             final long normalizedEndTimeMillis =
@@ -678,16 +682,27 @@ public class EditEventView implements View.OnClickListener, DialogInterface.OnCa
             if (normalizedEndTimeMillis < mModel.mStart) {
                 // mEnd should be midnight of the next day of mStart.
                 mModel.mEnd = mModel.mStart + DateUtils.DAY_IN_MILLIS;
+                modelEndTime.set(mModel.mEnd);
+                mEndTime.set(0, 0, 0, modelEndTime.getDay(), modelEndTime.getMonth(), modelEndTime.getYear());
+                mEndTime.normalize();
             } else {
                 mModel.mEnd = normalizedEndTimeMillis;
             }
+
+            mModel.mTimezone = Time.TIMEZONE_UTC;
+
+            // refresh UI to new start & end times
+            setDate(mStartDateButton, mStartTime.toMillis());
+            setTime(mStartTimeButton, mStartTime.toMillis());
+            setDate(mEndDateButton, mEndTime.toMillis());
+            setTime(mEndTimeButton, mEndTime.toMillis());
         } else {
             mStartTime.setTimezone(mTimezone);
             mEndTime.setTimezone(mTimezone);
             mModel.mStart = mStartTime.toMillis();
             mModel.mEnd = mEndTime.toMillis();
+            mModel.mTimezone = mTimezone;
         }
-        mModel.mTimezone = mTimezone;
         mModel.mAccessLevel = mAccessLevelSpinner.getSelectedItemPosition();
         // TODO set correct availability value
         mModel.mAvailability = mAvailabilityValues.get(mAvailabilitySpinner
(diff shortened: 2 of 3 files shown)


app/src/main/java/com/android/calendar/event/EditEventView.java
```diff
@@ -872,10 +890,37 @@ public void onCheckedChanged(CompoundButton buttonView, boolean isChecked) {
...
             mAllDayCheckBox.setChecked(true);
             // put things back in local time for all day events
+            // and force time at midnight
+            // also be robust against model having non-normalised all day event
+            // (start or end not midnight UTC or timezone not UTC), and force that
             mTimezone = Utils.getTimeZone(mActivity, null);
-            mStartTime.setTimezone(mTimezone);
-            mEndTime.setTimezone(mTimezone);
-            mEndTime.normalize();
+            {
```
Gitsaibot, line 897: Why are brackets here?

lmamane: To control (limit) the scope of variables year, month and day.

lmamane:
> With these changes my appointments and dates behave "strange" e.g. If I create a daily all day appointment the time span of 2 days is displayed in the partial view and not only the respective day.

I don't understand your bug description. What is "partial view"? I know of views "Day", "Weel", "Month" and "Agenda". I can reproduce your problem in none of these. What time zone are you in, maybe that has an influence? Do you have "use home time zone" enabled and in this case what is the system (OS-wide) time zone and what is the configured (in Etar) home time zone?

By "daily all day appointment", you mean an all-day appointment of duration 1 day (start day and end day are the same) that repeats daily?

lmamane:
> Click on a full day event in the month view and look at the time span.

When I click on a full day event in the month view, it brings me to the agenda view. If I click on the full-day event from the agenda view, I get to the detailed view, and indeed I can reproduce, but only if the time zone in use (the Etar-configured home time zone if its use is enabled, else the system time zone).

Weirdly, if I click on the same event from the day or week view, I cannot reproduce.

Now that I know how to reproduce, I'll try to take a look and fix this.

lmamane:
> With these changes my appointments and dates behave "strange" e.g. If I create a daily all day appointment the time span of 2 days is displayed in the partial view and not only the respective day.

Fixed. The functions Util.convertAllDayUtcToLocal and Util.convertAllDayLocalToUtc were plainly not working at all. They were correctly called from AgendaListView.onItemClick, but didn't do what they were supposed to.

app/src/main/java/com/android/calendar/month/SimpleWeekView.java
```diff
@@ -199,6 +201,14 @@ public SimpleWeekView(Context context) {
         initView();
     }
```
Gitsaibot, line 203: Why do we need a runnable to set the timezone?

app/src/main/java/com/android/calendar/Utils.java
```diff
@@ -793,8 +793,9 @@ public static long convertAlldayUtcToLocal(Time recycle, long utcTime, String tz
         }
         recycle.setTimezone(Time.TIMEZONE_UTC);
         recycle.set(utcTime);
-        recycle.setTimezone(tz);
-        return recycle.normalize();
+        Time target = new Time(tz);
```
Gitsaibot, line 796: I do not understand this change. Why is a new time object created here. The time is passed to the function and if not a new time object is created already.

app/src/main/java/com/android/calendar/month/SimpleWeekView.java
lmamane: So that the new timezone is taken into account when it is changed (in the settings).

app/src/main/java/com/android/calendar/Utils.java
lmamane: `recycle` (the time object "passed to the the function and if not created") contains the date of the all day event (with hour, minute and seconds set to zero, in other words time 00:00:00) in the UTC timezone.

`target` (the new time object created here) is set to the same date (still at time 00:00:00) but in the timezone `tz`.

The old code does:

```
recycle.set(utcTime);
recycle.setTimezone(tz);
```

I assume in the intention to achieve the same effect, but that does something different, which is the source of the bug being corrected here. Indeed, e.g. if `utcTime` is the timestamp (milliseconds since epoch) for e.g. 2023-08-28 00:00:00 UTC, and TZ is UTC-2 then recycle will contain THE SAME timestamp (the same "absolute" point in time), but expressed in timezone `tz`, that is 2023-08-27 22:00:00 UTC-2. That is not what this function should do. It should return 2023-08-28 00:00:00 UTC-2.

If you prefer, instead of creating `target` fresh each time, we can take __two__ arguments of `Time`, let's call them `recycle0` and `recycle1` and treat them as "object we can freely trash" as we now treat `recycle`

app/src/main/java/com/android/calendar/month/SimpleWeekView.java
lmamane: You will find exactly the same mechanism in `app/src/main/java/com/android/calendar/DayView.java`, with the same name, namely `mTZUpdater`

app/src/main/java/com/android/calendar/Utils.java
Gitsaibot: If it is as you describe why did it work before? I haven't looked deeper into the code and don't have that much time at the moment but it was displayed correctly before so something must have changed.

lmamane: It wasn't working before, not with all combinations of timezone and "use home time zone" on/off. See bugs #1173 #1216 #1164 #1208

Gitsaibot: I think the function does what the google programmer wanted. The question is why was it done that way? If you only wanted to change the timezone you could use this instead of setTimezone: 
switchTimezone(String timezone)

Convert this time object so the time represented remains the same, but is instead located in a different timezone.

lmamane: The comment at the top says:

> This (...) will adjust the time to be on a midnight boundary.

The old version clearly _does not adjust the time at all_, and specifically not _to be on a midnight boundary_. My change makes it do what the comment says it does. It adjusts the time to be on a midnight boundary.

One can discuss whether the comment or the code fits the "Google programmer's" intention better, and maybe you believe the code better expresses the intent of the programmer than the comment. However, my understanding of the general situation with all day events is:

- The date must be _stored_ as a UTC date on a midnight boundary.
- The UI code expects treats stuff in in system (or home when that option is enabled) timezone.
- So the stored timestamp must be converted to system/home timezone for the UI code, and the timestamp (date) that was changed in the UI must be converted to UTC in order to
- The above conversions are not "convert the date and time so that this is the same absolute point in time, expressed in a different timezone" but "take the same day-in-month number, the same month number and the same year number but slap system/home TZ, or UTC TZ respectively, on it"
- E.g. when the system/home timezone is UTC+2, the stored begin date of the all-day event is the number of milliseconds between the epoch and "2023-08-28 00:00:00 UTC+0", then the UI timestamp must be "2023-08-28 00:00:00 UTC+2".
- The methods `convertAlldayUtcToLocal` and `convertAlldayLocalToUTC` are supposed to do the conversions described above + force the time at midnight boundary (as per the comment) as a data sanitisation measure.


Generally, I think you are reading the documentation for `android.text.format.Time`, but the "Time" objects here are `com.android.calendarcommon2.Time`

Looking at the implementation of `switchTimezone` in that class on https://android.googlesource.com/platform/frameworks/opt/calendar/+/master/src/com/android/calendarcommon2/Time.java#210 shows it does not behave as your understanding of the documentation of `android.text.format.Time.switchTimeZone()`; here is the implementation:

```
public void switchTimezone(String timezone) {
        long msBefore = mCalendar.getTimeInMillis();
        mCalendar.setTimeZone(TimeZone.getTimeZone(timezone));
        mCalendar.setTimeInMillis(msBefore);
        mDstChangedByField = -1;
        readFieldsFromCalendar();
```

`mCalendar` is a `java.util.GregorianCalendar`; the call `mCalendar.setTimeInMillis(msBefore);` explicitly sets the time to the same absolute time (the same milliseconds since the epoch in UTC), see documentation of `java.util.GregorianCalendar` on https://developer.android.com/reference/java/util/GregorianCalendar, the method `setTimeinMillis` is inherited from `java.util.Calendar`, and the argument `long millis` is documented as:

millis&emsp;long: the new time in UTC milliseconds from the epoch.

So `switchTimezone` clearly does not do what is my understanding of the conversion that should happen there.

Gitsaibot: Right I remember there was a major change because the old API is deprecated. See 
https://github.com/Etar-Group/Etar-Calendar/commit/0b71aa9d424a8338ecdb977518e5d927c6aa07ca

Maybe it worked with the code before. I wonder if it is possible to change the time zone directly without going through a second Time object. But I didn't find a better solution at first sight to change the time zone in com.android.calendarcommon2.Time.

Gitsaibot (approved):

PR #1406, opened 2023-06-04
add missing dependency in external copyAAR

lmamane:
without that, fails to build for me, gradle errors out complaining of this missing dependency


jspricke (approved):

jspricke:
This makes `./gradlew build` work for me (tests are still failing, though).
@theimpulson would that be ok with you?

theimpulson (approved):

PR #1413, opened 2023-06-29
AlertReceiver: Return early if either context or intent is null

theimpulson:
## Summary

The intent and context can be null in a broadcast receiver. AlertReciever crashes the app if the intent is `null` for instance. This PR aims to fix that behavior by ensuring that all logic only runs if they are not null.

## Test

- Use the following command to send a broadcast without an intent action specified to crash Etar: `adb shell am broadcast -n ws.xsoh.etar.debug/com.android.calendar.alerts.AlertReceiver`
- Apply this patch and run the aforementioned command again
- Notice Etar no longer crashes


jspricke (approved):

jspricke:
Thanks!

PR #1421, opened 2023-07-18
allow editing calendar on existing events

### LINKED ISSUE #62: Move event between calendars
Firstly, thanks for your work.
Please, could you add support to move events between calendars.
Thanks again.

### LINKED ISSUE #101: Option to change calendar for added events missing
In both S Planner and Etar you can choose what Calendar an Event should use, but only in S Planner it is possible to modify the Calendar once it have been added.

### LINKED ISSUE #812: Can no longer change calendar?
I came to report that in the last 1-2 weeks, I lost the ability to edit the calendar of an existing event.
To my surprise, it seems this feature isn't even present in Etar, but has been requested.

However, I swear I have been using this feature for the last few months?
I only lost the ability in the last couple of weeks but I'm not sure which of the recent updates from DAVx5 or Etar that stopped it working.

I use Etar, DAVx5 & Nextcloud....previously in Etar, I could tap an _event > edit > tap calendar name > drop down_ menu to select the calendar I want to change the event to.

Now I just feel crazy???

Huawei LYA-L29
Android 10
Etar 1.0.23

### LINKED ISSUE #1134: Colour change of single events does not work
After creating an event, when I try to change colour from default and click on the colour pallet button, there shows a blank loading window and nothing more.

Issue on: Android 9 PKQ1.181021.001 with MIUI Global 11.0.11

### LINKED ISSUE #1266: Not possible to change calendar on an existing event
With Android 12 and Etar 1.0.30 it is not possible to change calendar on an existing event. The only solution is to delete it and re-create it.

### LINKED ISSUE #1351: Color picker for event color is not working
When click on palette icon, waiting circle shows and rotates and rotates...
Screenshot_20230302-195630

### LINKED ISSUE #1400: Change calendar
Once I have entered an event, it is impossible to change the calendar it belongs to.  I can change everything else but not the calendar.

### LINKED ISSUE #1410: Feature request: move event to other calendar
Unfortunately, we have to maintain several agendas i. e. several calendars sync'ed via DAVX5. Sometimes it happens that an event is opened by the user in a wrong calendar.

Etar does not allow to move this event to another calendar. So either the event must be deleted and copied/re-typed into the correct calendar or users use their Laptop app (in our case Thunderbird).

jonas-haeusler:
these changes have been lying around for some time, finally found the time to finalize them and create this merge request.

in essence, this will just delete the old event and recreate it with a new id in the new calendar.
please test thoroughly and let me know if something is unclear.

fixes a bunch of open (duplicated) issues, here a few i found while quickly scanning the backlog:
fixes #62 
fixes #101
fixes #812
fixes #1266 
fixes #1400
fixes #1410

and also fixes #1351 and fixes #1134: the color picker is now hidden if the calendar does not provide any colors.

### DIFF
--- a/app/src/main/java/com/android/calendar/event/EditEventFragment.java
+++ b/app/src/main/java/com/android/calendar/event/EditEventFragment.java
@@ -734,11 +734,31 @@ public class EditEventFragment extends Fragment implements EventHandler, OnColor
                         setModelIfDone(TOKEN_REMINDERS);
                     }
 
+                    // disable non-synced calendars for recurring events
+                    // disable all calendars for recurring events and sdk<30
+                    final String selection;
+                    final String[] selectionArgs;
+                    if (!TextUtils.isEmpty(mModel.mRrule)
+                            && Build.VERSION.SDK_INT < Build.VERSION_CODES.R) {
+                        // recurring event, api level < 30. disable changing calendars.
+                        selection = EditEventHelper.CALENDARS_WHERE;
+                        selectionArgs = new String[] { Long.toString(mModel.mCalendarId) };
+                    } else if (!TextUtils.isEmpty(mModel.mRrule)) {
+                        // recurring event, api level >= 30. enable changing calendars to synced calendars.
+                        selection = EditEventHelper.CALENDARS_WHERE_SYNCED_WRITEABLE_VISIBLE;
+                        selectionArgs = null;
+                    } else {
+                        // non recurring event. enable changing calendars to all calendars.
+                        selection = EditEventHelper.CALENDARS_WHERE_WRITEABLE_VISIBLE;
+                        selectionArgs = null;
+                    }
+
                     // TOKEN_CALENDARS
                     mHandler.startQuery(TOKEN_CALENDARS, null, Calendars.CONTENT_URI,
                             EditEventHelper.CALENDARS_PROJECTION,
-                            EditEventHelper.CALENDARS_WHERE_WRITEABLE_VISIBLE,
-                            null /* selection args */, null /* sort order */);
+                            selection,
+                            selectionArgs,
+                            null /* sort order */);
 
                     // TOKEN_COLORS
                     mHandler.startQuery(TOKEN_COLORS, null, Colors.CONTENT_URI,
--- a/app/src/main/java/com/android/calendar/event/EditEventFragment.java
+++ b/app/src/main/java/com/android/calendar/event/EditEventFragment.java
@@ -738,11 +738,11 @@ public class EditEventFragment extends Fragment implements EventHandler, OnColor
                     final String[] selectionArgs;
                     final boolean isRecurring = !TextUtils.isEmpty(mModel.mRrule);
                     if (isRecurring && Build.VERSION.SDK_INT < Build.VERSION_CODES.R) {
-                        // recurring event, api level < 30. disable changing calendars.
+                        // recurring event AND api level < 30. disable changing calendars.
                         selection = EditEventHelper.CALENDARS_WHERE;
                         selectionArgs = new String[] { Long.toString(mModel.mCalendarId) };
                     } else if (isRecurring) {
-                        // recurring event, api level >= 30. enable changing calendars to synced calendars.
+                        // recurring event AND api level >= 30. enable changing calendars to synced calendars.
                         selection = EditEventHelper.CALENDARS_WHERE_SYNCED_WRITEABLE_VISIBLE;
                         selectionArgs = null;
                     } else {
--- a/app/src/main/java/com/android/calendar/event/EditEventFragment.java
+++ b/app/src/main/java/com/android/calendar/event/EditEventFragment.java
@@ -308,7 +308,7 @@ public class EditEventFragment extends Fragment implements EventHandler, OnColor
         super.onAttach(activity);
         mActivity = (AppCompatActivity) activity;
 
-        mHelper = new EditEventHelper(activity, null);
+        mHelper = new EditEventHelper(activity);
         mHandler = new QueryHandler(activity.getContentResolver());
         mModel = new CalendarEventModel(activity, mIntent);
         mInputMethodManager = (InputMethodManager)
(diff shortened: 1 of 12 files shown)


Luensche:
After a quick test: Everything looks fine. :)

dumblob:
Is this transactional? I.e. using two-phase commit (i.e. first creating a copy in the destination calendar, and if successful, then removing the original).

jonas-haeusler:
no, currently, the existing event is deleted and then the new event is inserted. should the deletion fail, the event is lost. i agree that this is suboptimal, and first inserting and only deleting on success would already provide a safety net. i don't know how easy (or not) this would be to implement. sadly, we can't just slap @Transactional on here :-(

on another note: changing the calendar on a recurring event seems to be broken right now. this will require some more work...

BalooRJ1:
Thank you for implementing this! Looking forward to seeing it pushed to the final version.

Cwpute:
> Thank you for the pull request. I think many have been waiting for this ;).
> 
> @ALL It would help us if many test these changes. The more test it in advance the faster we can create a new version. Here is the link to the apk: https://github.com/Etar-Group/Etar-Calendar/actions/runs/5603553145

Tested on Android 11, Nokia TA-1157: it works too!

mgw2013:
Happy to test but would need a link.

kerbless:
> in essence, this will just delete the old event and recreate it with a new id in the new calendar.

Is this really the only way to do this?

jonas-haeusler:
recurring events should currently work okay-ish (i think there were some problems with recurrence exceptions when moving them to local/synced calendars) and the whole thing is now transactional, so all or nothing, which is nice.

i currently don't know how to proceed with the todo [here](https://github.com/Etar-Group/Etar-Calendar/pull/1421/files#diff-82408967c5cac1e46e20874717a475ffcc3ab75668e6f0c0125043c96f94ccd8R674), as this method is not available before api level 30 and there seems to be no easy workaround for that. maybe the moving of recurring events should just be disabled for now (or for users of api level <30) to get this merged?

jspricke:
Enabling it for API 30 only would be fine with me.

mgw2013:
F-Droid still V 39 which doesn't install in A14 LOS  variant.

myshen:
Tested on Android 14, Samsung Galaxy. works great!

jonas-haeusler:
> Ready to review?

sorry for the late reply. yes, this is ready for review and testing. if possible, I would appreciate if you could test moving events between calendars (of different providers if possible, and local/synced calendars) and check whether all event properties stay as expected. thank you!

mgw2013:
Would love but can't locate the .apk for testing purposes. Might be me or this mobile client...

jonas-haeusler:
> Apart from the two things mentioned above, I haven't noticed anything else so far. Can you have a look? It would be cool if we could complete the PR before it's one year old ;)

@Gitsaibot these issues should now be fixed (amongst a few others). if you find the time, please retest and let me know whether it is working for you.

i hope that recurrences are mostly working now. they were a lot harder to get right than i expected.
in fact, one case that is currently still broken is if a recurrence is synced with the EXDATE property set, and then the start time of the recurrence is changed by Etar, then this change does not currently update the start time within the EXDATE property, effectively removing any exceptions defined by EXDATE. however, adding EXDATE support is out of scope for this imo.
 

> Would love but can't locate the .apk for testing purposes. Might be me or this mobile client...

@mgw2013 see https://github.com/Etar-Group/Etar-Calendar/actions/runs/9357933167.

mgw2013:
Just tried to move a single instance of a recurring event to another calendar - doesn't work. The drop down menu is greyed out. 

However, moving the entire series to another calendar works, which is much appreciated! -mgw

mgw2013:
Uhhh....it looks like the changes are not replicated via WebDav to the host...waited for several minutes and issues several sync requests on all devices. -mgw

app/src/main/java/com/android/calendar/event/EditEventFragment.java
```diff
@@ -734,13 +734,29 @@ protected void onQueryComplete(int token, Object cookie, Cursor cursor) {
...
 
+                    final String selection;
+                    final String[] selectionArgs;
+                    final boolean isRecurring = !TextUtils.isEmpty(mModel.mRrule);
+                    if (isRecurring && Build.VERSION.SDK_INT < Build.VERSION_CODES.R) {
+                        // recurring event AND api level < 30. disable changing calendars.
+                        selection = EditEventHelper.CALENDARS_WHERE;
+                        selectionArgs = new String[] { Long.toString(mModel.mCalendarId) };
+                    } else if (isRecurring) {
+                        // recurring event AND api level >= 30. enable changing calendars to synced calendars.
```
Gitsaibot, line 745: Why is this only activated for sync calender?

jonas-haeusler: local, non-synced calendars don't allow recurrence exceptions. so to be on the safe side and not to lose any events, I disabled moving recurrences to non-synced calendars.

Gitsaibot: Ah okay, maybe a short comment here why it's disabled would be helpful.

jonas-haeusler:
> Just tried to move a single instance of a recurring event to another calendar - doesn't work. The drop down menu is greyed out.

yeah, I disabled moving single instances out of recurrences, as this would be a lot more work. its currently either all or nothing.

> it looks like the changes are not replicated via WebDav to the host

can you please tell me how your setup looks like? which sync provider do you use on android and which caldav server software are you running?

mgw2013:
> can you please tell me how your setup looks like? which sync provider do you use on android and which caldav server software are you running?
Android 14 May w/ sec update DAVx5 from F-Droid. Calendar on GMX mail system - not sure what they r using.

Anyhow - I did a new test - I just takes a longer time to replicate the changes to the host an vice versa. So things r working as designed but slow.

Gitsaibot (approved):

szaimen:
🎉🎉🎉🎉🎉

mgw2013:
Good job! Txs for ur effort - much appreciated

wilhelmy:
Awesome, thank you @jonas-haeusler!

Issue #1443, opened 2023-09-12
 java.lang.IllegalArgumentException


Gitsaibot:
`Exception java.lang.IllegalArgumentException: Object returned from onCreateLoader must not be null
  at androidx.loader.app.LoaderManagerImpl.createAndInstallLoader (LoaderManagerImpl.java:385)
  at androidx.loader.app.LoaderManagerImpl.initLoader (LoaderManagerImpl.java:421)
  at com.android.calendar.month.MonthByWeekFragment.onActivityCreated (MonthByWeekFragment.java:318)
  at androidx.fragment.app.Fragment.performActivityCreated (Fragment.java:2996)
  at androidx.fragment.app.FragmentStateManager.activityCreated (FragmentStateManager.java:580)
  at androidx.fragment.app.FragmentStateManager.moveToExpectedState (FragmentStateManager.java:285)
  at androidx.fragment.app.FragmentStore.moveToExpectedState (FragmentStore.java:112)
  at androidx.fragment.app.FragmentManager.moveToState (FragmentManager.java:1647)
  at androidx.fragment.app.FragmentManager.dispatchStateChange (FragmentManager.java:3128)
  at androidx.fragment.app.FragmentManager.dispatchActivityCreated (FragmentManager.java:3072)
  at androidx.fragment.app.FragmentController.dispatchActivityCreated (FragmentController.java:251)
  at androidx.fragment.app.FragmentActivity.onStart (FragmentActivity.java:502)
  at androidx.appcompat.app.AppCompatActivity.onStart (AppCompatActivity.java:251)
  at android.app.Instrumentation.callActivityOnStart (Instrumentation.java:1457)
  at android.app.Activity.performStart (Activity.java:8259)
  at android.app.ActivityThread.handleStartActivity (ActivityThread.java:3776)
  at android.app.servertransaction.TransactionExecutor.performLifecycleSequence (TransactionExecutor.java:221)
  at android.app.servertransaction.TransactionExecutor.cycleToPath (TransactionExecutor.java:201)
  at android.app.servertransaction.TransactionExecutor.executeLifecycleState (TransactionExecutor.java:173)
  at android.app.servertransaction.TransactionExecutor.execute (TransactionExecutor.java:97)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:2311)
  at android.os.Handler.dispatchMessage (Handler.java:111)
  at android.os.Looper.loopOnce (Looper.java:238)
[...]

Issue #1439, opened 2023-09-09
App crashes on start 


Outlet2048:
Just updated to the latest version and whenever I open the app for the first time it crashes
Opening it again works just fine

Until the app gets killed in the background or something and then it repeats

Oneplus9pro
Android 13

psrunner71:
I also just updated Android and when I open the app it crashes.

Lenovo Tab M10 Plus 3rd Gen (TB125FU)
Android 13

lucyferu:
> Opening it again works just fine

same here, but reopening never works.
the app crashes always.

Gitsaibot:
A debug log would be helpful.

theimpulson:
> @theimpulson I think this has something to do with your migration to androidx fragment library. Do you have an idea?
> 
> Maybe same as: https://stackoverflow.com/questions/60175687/loadermanager-loadercallbacks-android-studio

I will check

Gitsaibot:
@all Would be nice if you could test this apk: https://github.com/Etar-Group/Etar-Calendar/actions/runs/6181056696

lucyferu:
> @ALL Would be nice if you could test this apk: https://github.com/Etar-Group/Etar-Calendar/actions/runs/6181056696

it works just fine 😸

JanaKasparova:
Sorry, keeps crashing for me. I can see a daily schedule for a while but that's all :( I have two version installed: 1.0.30 and the above mentioned version. I do not know if that matters.

lucyferu:
> Sorry, keeps crashing for me. I can see a daily schedule for a while but that's all :( I have two version installed: 1.0.30 and the above mentioned version. I do not know if that matters.

I also have both installed, but the testing one works for me nonetheless.

0verEngineer:
[...]

PR #1444, opened 2023-09-14
MonthByWeekFragment: Don't init loader in certain cases

theimpulson:
## Summary

The new androidx LoaderManager requires it to be NonNull. Strip out the cases that currently return null and add them as a check before initializing the LoaderManager.

Also, switch to androidx methods to get the LoaderManager instance.

## Issues

- #1439
- #1443


theimpulson:
@Gitsaibot could you please check if this fixes your issue?

Gitsaibot (approved):

theimpulson:
@Gitsaibot fine to merge?

PR #1461, opened 2023-10-03
when user changes event timezone, start/end same time in new timezone

lmamane:
as opposed to silently and invisibly converting the absolute datetime to new timezine.

E.g. if the event was:

start: 14 February 19:00 (invisibly: UTC+2)
end:   14 February 23:00 (invisibly: UTC+2)
tz:    UTC+2

and the users changes the tz to UTC+3, the times must be changed to:

start: 14 February 19:00 (invisibly: UTC+3)
end:   14 February 23:00 (invisibly: UTC+3)

as opposed to remaining:

start: 14 February 19:00 (invisibly: UTC+2)
end:   14 February 23:00 (invisibly: UTC+2)

and thus becoming on save:

start: 14 February 20:00 UTC+3
end:   15 February 00:00 UTC+3

which is not what the user meant...

### DIFF
--- a/app/src/main/java/com/android/calendar/Utils.java
+++ b/app/src/main/java/com/android/calendar/Utils.java
@@ -2246,4 +2246,18 @@ public class Utils {
         }
     }
 
+    /**
+     * Change a Time object to be the same (year, month, day, hour, minute, second) tuple
+     * but in another timezone
+     *
+     * @param t The Time object to modify
+     * @param timezone the new timezone
+     */
+    public static void changeTimezoneOnly(Time t, String timezone) {
+        Time pivot = new Time(timezone);
+        pivot.set(t.getSecond(), t.getMinute(), t.getHour(),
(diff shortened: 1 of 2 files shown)


jspricke:
Thanks! Sounds good to me but did not test.

lmamane:
Choose two DIFFERENT timezones that have a DIFFERENT offset to UTC. Let's cal them TZ1 and TZ2, let's assume TZ1 is UTC+5 and TZ2 is UTC+1.

Create an event in timezone TZ1, e.g. from 13:00 to 14:00 save it; you must exit the details (event edition) screen.

Edit the event again.

Change timezone to TZ2.

Do NOT TOUCH the times or dates of the event in ANY WAY.

Save.

Look at the event. Notice it is saved as 9:00 to 10:00 in TZ2 but should have been 13:00 to 14:00 in TZ2.

Gitsaibot (approved):

PR #1469, opened 2023-10-06
gh#1463 fix wrong day display in widget

lmamane:
The bug ultimately came from the fact that:
 - convertAllDayLocalToUtc and convertAllDayUtcToLocal are not guaranteed to preserve any part of the value of their recycle argument
 - older implementation of convertAlldayUtcToLocal just happened to accidentally preserve recycle's timezone
 - newer implementation does not; note that convertAlldayLocalToUtc already didn't preserve recycle's timezone previously
 - the code in buildFromCursor assumed that recycle was preserved, in particular by passing it to populateDayInfo while the latter critically depends on recycle having the correct timezone


Haisha10 (approved):

jspricke:
Thanks! This looks ok to me, but I did not test it.

Gitsaibot (approved):

PR #1470, opened 2023-10-06
review all uses of convertAlldayUtcToLocal

lmamane:
this pull request is one commit above #1469; it tries to proactively fix the same underlying issue everywhere convertAlldayUtcToLocal is used in a suspicious way.


lmamane:
Since https://github.com/Etar-Group/Etar-Calendar/pull/1469 was applied, rebase to that this PR shows as one commit instead of showing commits that are already applied.

PR #1472, opened 2023-10-09
Android 14 support

theimpulson:
## Summary

- Bump libraries to the latest stable release (AOSP builds need this)
- Specify foreground service types, Use FOREGROUND_SERVICE_TYPE_DATA_SYNC from Android 10 to 13, FOREGROUND_SERVICE_TYPE_SYSTEM_EXEMPTED from Android 14
- Bump compile and target SDK to 14

## Test

New install > Grant all permissions > create new event in an offline calendar > Notice event notification appears and Etar doesn't crashes

### DIFF
--- a/app/src/main/java/com/android/calendar/Utils.java
+++ b/app/src/main/java/com/android/calendar/Utils.java
@@ -1787,7 +1787,7 @@ public class Utils {
         filter.addAction(Intent.ACTION_LOCALE_CHANGED);
 
         CalendarBroadcastReceiver r = new CalendarBroadcastReceiver(callback);
-        c.registerReceiver(r, filter);
+        ContextCompat.registerReceiver(c, r, filter, ContextCompat.RECEIVER_NOT_EXPORTED);
         return r;
     }
 
--- a/app/src/main/java/com/android/calendar/Utils.java
+++ b/app/src/main/java/com/android/calendar/Utils.java
@@ -21,6 +21,7 @@ import static android.provider.CalendarContract.EXTRA_EVENT_BEGIN_TIME;
 import android.Manifest;
 import android.accounts.Account;
 import android.app.Activity;
+import android.app.AlarmManager;
 import android.app.PendingIntent;
 import android.app.SearchManager;
 import android.content.BroadcastReceiver;
@@ -54,6 +55,7 @@ import android.util.Log;
 import android.widget.Toast;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.RequiresApi;
 import androidx.appcompat.widget.SearchView;
 import androidx.core.content.ContextCompat;
 import androidx.core.graphics.ColorUtils;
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/android/calendar/Utils.java
```diff
@@ -1787,7 +1787,7 @@ public static BroadcastReceiver setTimeChangesReceiver(Context c, Runnable callb
         filter.addAction(Intent.ACTION_LOCALE_CHANGED);
 
         CalendarBroadcastReceiver r = new CalendarBroadcastReceiver(callback);
-        c.registerReceiver(r, filter);
+        ContextCompat.registerReceiver(c, r, filter, ContextCompat.RECEIVER_NOT_EXPORTED);
```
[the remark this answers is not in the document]
Gitsaibot: I get the same message. But at first glance I do not see what should be wrong.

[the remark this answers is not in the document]
Gitsaibot: After several gradle runs and changing the package versions back and forth, the error is gone for me.

app/src/main/java/com/android/calendar/alerts/AlertReceiver.java
```diff
@@ -139,7 +139,11 @@ public static void beginStartingService(Context context, Intent intent) {
 
             if (Utils.isMOrLater()) {
                 if (pm.isIgnoringBatteryOptimizations(context.getPackageName())) {
-                    context.startService(intent);
+                    if (Utils.isUpsideDownCakeOrLater() && Utils.canScheduleAlarms(context)) {
```
Gitsaibot, line 142: Why do we need the if else here if  **context.startService(intent);**  is always executed?

theimpulson: Oh good catch, should return if permission isn't given.

theimpulson: Done

app/src/main/java/com/android/calendar/alerts/AlertReceiver.java
```diff
@@ -132,14 +137,17 @@ public static void beginStartingService(Context context, Intent intent) {
...
-            if (pm.isIgnoringBatteryOptimizations(context.getPackageName())) {
-                if (Utils.isOreoOrLater()) {
-                    context.startForegroundService(intent);
-                } else {
+            if (Utils.isMOrLater()) {
+                if (pm.isIgnoringBatteryOptimizations(context.getPackageName())) {
+                    if (Utils.isUpsideDownCakeOrLater() && !Utils.canScheduleAlarms(context)) {
+                        return;
+                    }
                     context.startService(intent);
```
Gitsaibot, line 145: Why is no "context.startForegroundService(intent);" needed here anymore?

theimpulson: `AlertService` calls `startForeground` Android 8.0+. In the case of Android 10.+, its called as well but with a specific type.

Gitsaibot: Maybe I understand something wrong but I thought from android 8 you should not use startService. This is also described here: https://www.hellsoft.se/your-guide-to-foreground-services-on-android/

theimpulson: `startService` basically starts a service without promising anything, whether foreground or background. If you call `startForeground` within 5 seconds, the service becomes Foreground and it doesn't crash. On the other hand, `startForegroundService` implicitly promises that the service will definitely call `startForeground`. 

Regardless, I have changed it back to how it was. The only thing changed now is it has more `if..else` loops.

jspricke:
Thx!

PR #1476, opened 2023-10-19
Material3 Migration & Misc. Improvements

theimpulson:
## Summary

This PR migrates the theme setup to use Material3 themes and does misc—improvements to the overall setup.

- Switch to Material3 theme
- Drop custom attributes altogether (resources are now accessed directly, tint and colors are accessed from standard public APIs)
- Switch to MaterialAlertDialog everywhere possible
- Switch to MaterialDatePicker for "Goto" feature

## Dependent PRs

- https://github.com/Etar-Group/standalone-calendar-timezonepicker/pull/7


theimpulson:
> Is it still possible to change the view between light dark and black with your changes? For me, the view no longer changes when I choose between light and black, for example.

Still working on that, but will try to keep it possible.

> I had a quick look and I think the edit event page needs some work to look like it did before.

Fixed, thanks for checking. I assume there are more areas that need fixes as I can see multiple activities that still have old themes and behavior.

Bnyro:
This looks pretty cool, nice work!

I have some suggestions for minor additional changes that would make everything look a bit more MD3 :)
1. Perhaps we want to use the new [`MaterialSwitch`](https://m3.material.io/components/switch/overview) instead of the old MD3 one (it's possible to just replace the import everywhere to the one of the material design library, it's a drop-in replacement).
2. The "Add reminder" button doesn't look very well in place in my opinion. I'd suppose using a normal `MaterialButton` with the `OutlinedButton` style to make it fit in better.
3. The color picker dialog is still using Material Design 2 (I guess that's just because the app uses an external library for that, I'm not familiar with the code unfortunately. If so, this point can be ignored).
4. The preference dialogs can be migrated to MD3 without much efforts too, I've implemented this for another app [here](https://codeberg.org/Gaukler_Faun/FOSS_Browser/pulls/1025/files). 

If you have any questions let me know. I'm very excited to see this ongoing effort to move the MD3 :)

Cheers and keep on the good work ^)

theimpulson:
@Gitsaibot a new review would be good

theimpulson:
> The "Primary color" setting does not work.

@Gitsaibot where is this setting? Do you mean Themes?

app/src/main/java/com/android/calendar/AllInOneActivity.java
```diff
@@ -926,18 +923,19 @@ public boolean onOptionsItemSelected(MenuItem item) {
                 t = todayTime;
             }
 
-            DatePickerDialog.OnDateSetListener datePickerListener = new DatePickerDialog.OnDateSetListener() {
```
Gitsaibot, line 929: Can you move the part with the DatePicker to a separate PR? I'm currently looking at this myself because of "GoTo errors" Also, small parts can be checked more quickly and transferred to the main branch. What about lines 917-924, they can also be removed, can't they?

theimpulson: @Gitsaibot last night I thought splitting the date and time picker to separate PR would be a good idea but just now I remembered that they are Material classes and the current source isn't using the Material theme at all. I feel that might just introduce more weird bugs.

Gitsaibot: Yes, that thought occurred to me too. But if these are minor visual bugs, I don't think that would be such a big problem. And I don't think it will be that easy to change everything at once.

theimpulson: Lines removed.

theimpulson:
One of the issues I am not sure why is happening is that MaterialAlertDialogs are not visible unless they are made visible by calling `show` explicitly on the dialog instance (not the dialog fragment instance) which is weird. Seems to be only happening on Picker classes though.

theimpulson:
@Gitsaibot I have restricted pure black to only apply if you select the dark or black theme. Maybe you selected theme first then enabled pure black setting?

theimpulson:
> I don't quite understand that. The "pureBlack" setting should only has an effect if "System standard" is selected as the theme.
> 
> Step 1 choose systemtheme as theme Step 2 activate pureBlack setting
> 
> This is what I wanted to show with the screenshots

Got it, I was confused about when it should apply. I will fix it shortly.

theimpulson:
Rebased and fixed the pure black theme issue.

theimpulson:
> 2\. The purple switch see screenshot(Comment 16 January) above. It is still purple.

I would like to fix dialogs in other PRs as they seem to be behaving weirdly for some unknown reason. I assume they will need a bit of work and cleanup.

theimpulson:
> > Rebased and fixed the pure black theme issue.
> 
> Now the black theme has the same background as the dark theme. Or am I wrong ?

My bad, fixed

theimpulson:
I wasn't expecting these many issues. Let's hold the PR for now then. I will rework and open it again once ready.

PR #1483, opened 2023-10-27
TimeReminder_Partial_fix, add more fixed time into the list. The fund…

Jennie-pro:
TimeReminder_Partial_fix to add more fixed times to the reminder list to satisfy users. The underlying issue needs to be addressed further.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -105,6 +105,7 @@ dependencies {
     implementation 'androidx.appcompat:appcompat:1.6.1'
     implementation 'androidx.constraintlayout:constraintlayout:2.1.4'
     implementation 'com.google.android.material:material:1.8.0'
+    implementation project(path: ':external:calendar')
     testImplementation 'junit:junit:4.13.2'
 
     coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:2.0.3'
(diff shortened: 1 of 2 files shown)


Jennie-pro:
As far as the idea of building this project is concerned, the time reminder is read via an array, not a custom one. I think a more efficient way to do this in the short term would be to add more options. Of course, if you have more time and energy, you can change the whole structure of the time reminder. What do you think?

app/build.gradle
```diff
@@ -105,6 +105,7 @@ dependencies {
     implementation 'androidx.appcompat:appcompat:1.6.1'
     implementation 'androidx.constraintlayout:constraintlayout:2.1.4'
     implementation 'com.google.android.material:material:1.8.0'
+    implementation project(path: ':external:calendar')
```
jspricke, line 108: why do you need this?

jspricke (commented):
I also think that the list is already too long and would rather opt for a numerical filed and/or a slider.

Jennie-pro:
Yeah, I agree that the idea of using a numerical filed and/or a slider is good. However, it takes more time and effort changing the structure of time reminder. What do you think?

jspricke:
I think we can wait till someone implements that, then. So closing.

app/build.gradle
Jennie-pro: aha, I have no idea why it changes. I did not make any change over there. :((

Issue #1497, opened 2023-11-02
Enable automatic per-app language support


ploink:
This is a new feature for android 13+ that allows one to pick an app's language from system settings or app info. 
I find this very useful as I have the android system in English but prefer some apps like Etar in my native language.

Implementation is very easy:
https://developer.android.com/guide/topics/resources/app-languages#auto-localeconfig
I have tried it locally and it works.

jspricke:
Could you send this as a pull request?

ploink:
> Could you send this as a pull request?

Sorry no luck. It appears I was using my older fork. I just cloned the most recent code and Studio wont even sync the project. And I would need to read the manual on how to create a pull request. I am not a developer.

It is very simple:

```
$ echo "unqualifiedResLocale=en" > app/src/main/res/resources.properties
$ git add app/src/main/res/resources.properties
```

Then add this to the module-level build.gradle:

```
    android {
      androidResources {
        generateLocaleConfig true
      }
    }
```

I hope you or someone else can pick this up. Thanks.

jspricke:
@ploink I have opened #1502 can you have a look?

PR #1502, opened 2023-11-05
Enable automatic per-app language support (Closes: #1497)

jspricke:
Thanks to ploink


ploink:
Thanks jspricke, it works perfectly.

PR #1556, opened 2024-01-11
New version 1.0.39

Gitsaibot:


metadata/en-US/changelogs/39.txt
```diff
@@ -0,0 +1 @@
+- Small bugfixes
```
jspricke, line 1: maybe be a bit more specific, like: "Fix crash on month and week view due to invalid rrule"?

Either way, please go ahead with the release :).

jspricke (approved):

Issue #1450, opened 2023-09-21
Background Notification Dialog Nags User


landry314:
I never use notifications and i dont want Etar running in the background.

You have a dialog to remind users that Etar can run in the background and it says you can allow or deny this. So I deny it and then I go to App settings and completely deny it there.

Every time I open the app, Etar asks me again if i want to let Etar run in the background. No, I don't. Why is it nagging me?

Also, the (i) at the top waits there for me to tap it by accident.

When you make a choice about a setting it should make the choice and not bother you about your choice.

In issue #1346 this is discussed a bit and you say you SHOULD have Etar run in the background or you should stop using the app. While I completely disagree with this, I must ask, why would you ask the user what they want then push them in one direction and nag them until they give in? Why ask the user in the first place? Why not just require background access to use the app and leave it at that? Again, I don't need or want this unnecessary feature but why would you nag the user?

And that (i) needs to be moved off the top bar into the settings or maybe the menu which i think is still weird because it isnt a function, it is a setting.

I am using LineageOS 20.0 on a Pixel 5. Rooted.

quite:
+1 here. I would use Etar to look at the calendar, I never want it to run in the background, or display any notifications. With the current nagging, Etar is really not usable.

kajdo:
same here ... i think a simple checkbox in the setting "don't check battery optimization" with the remark "reminders will not work reliably" / "I know what I'm doing" would do the trick

i like the app a lot, but considering that most users might be on the track for customizing their devices (otherwise they would possibly use their stock calendar anyway) - I think an option ... maybe even a bit hidden in an "advanced" or "danger" section could solve this issue for good and everybody is happy

if the setting is activated - skip the check at the start of the app - that's it ... plenty of issues and discussions ended, and the dev team don't have to be the ones who are responsible if issues are opened because of lag in reminders/notifications if this setting is done

just a thought

Gitsaibot:
> if the setting is activated - skip the check at the start of the app - that's it ... plenty of issues and discussions ended, and the dev team don't have to be the ones who are responsible if issues are opened because of lag in reminders/notifications if this setting is done

We are always responsible for everything. It would be a dream if it weren't so. 

It has to be said that the setting also causes certain background processes to crash. Google logs this and it is then displayed in the PlayConsole, which leads to a poor rating/visibility of the app in the Playstore. 

[...]

PR #1570, opened 2024-01-22
add pref that user does not want a background service (#1450)

mweller83:
Adds an option in general preferences to disable showing a "Background Notification Dialog", see https://github.com/Etar-Group/Etar-Calendar/issues/1450

### DIFF
--- a/app/src/main/java/com/android/calendar/AllInOneActivity.java
+++ b/app/src/main/java/com/android/calendar/AllInOneActivity.java
@@ -425,8 +425,8 @@ public class AllInOneActivity extends AbstractCalendarActivity implements EventH
 
     private void checkAndRequestDisablingDoze() {
         if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) return;
-
-        if (!dozeDisabled()) {
+        boolean neverEnableDoze = Utils.getSharedPreference(getApplicationContext(), GeneralPreferences.KEY_NEVER_ENABLE_DOZE, false);
+        if (!dozeDisabled() && !neverEnableDoze) {
             Intent intent = new Intent();
             intent.setAction(Settings.ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS);
             intent.setData(Uri.parse("package:" + getApplicationContext().getPackageName()));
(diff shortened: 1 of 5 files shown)


app/src/main/res/xml-v26/general_preferences.xml
```diff
@@ -128,6 +128,11 @@
             app:fragment="com.android.calendar.settings.QuickResponsePreferences"
             app:title="@string/quick_response_settings"
             app:summary="@string/quick_response_settings_summary" />
+
+        <SwitchPreference
```
Gitsaibot, line 132: Please move to the experimental preference section

app/src/main/java/com/android/calendar/AllInOneActivity.java
```diff
@@ -425,8 +425,8 @@ private void checkAppPermissions() {
 
     private void checkAndRequestDisablingDoze() {
         if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) return;
-
-        if (!dozeDisabled()) {
+        boolean neverEnableDoze = Utils.getSharedPreference(getApplicationContext(), GeneralPreferences.KEY_NEVER_ENABLE_DOZE, false);
```
Gitsaibot, line 428: I find the name of the setting totally confusing and I'm not sure if the users understand it correctly. When I activate the setting but battery optimization is activated in the phone settings, no dialog appears. Why not "Don't check battery optimization" ?

app/src/main/res/xml-v26/general_preferences.xml
```diff
@@ -128,6 +128,11 @@
...
+            app:key="pref_never_enable_doze"
+            app:defaultValue="false"
+            app:title="@string/preferences_never_enable_doze" />
```
Gitsaibot, line 135: I think there should be a short description for this setting, e.g. use at your own risk! Notifications will no longer work, widget will no longer update and background services will crash.

mweller83:
Thanks for your comments! So I moved & renamed the pref, and added a little summary like you suggested.

Gitsaibot (approved):

Issue #1138, opened 2022-04-06
Mute alarms from specific calendar


Kalytis:
Nextcloud's birthday calendar pops up a notification at midnight. It's pissing off, especially on my phone.

I was so sure there was a way to disable alarms for a whole calendar. I thought I had no alarms for my birthday calendars before, but since the last update, I have them again, and can't find a way to block them.

Thanks a lot for your help.

Zocker1999NET:
might be related to #1107

niccokunzmann:
I think, this is a duplicate of #1107.

akki42:
This has been requested / discussed for years. Please see https://github.com/Etar-Group/Etar-Calendar/issues/67 and https://github.com/Etar-Group/Etar-Calendar/issues/783.

Issue #783, opened 2020-09-29
Feature request: allow disabling reminders per-calendar


runejuhl:
(I looked through the issues and didn't find anything similar; hope I didn't miss anything)

I have a bunch of calendars that are shared, and one of the things that quickly becomes an issue is when I'm interested in seeing calendar events from shared calendars and these have reminders set. This means that I often get a bunch of reminders for events that I have no need to be reminded for.

Would it be possible to add a toggle for disabling reminders per-calendar? It could also be part of DAVx5/other synchronization software, but I feel that it'd be better to have the feature in Etar.

ploink:
I have exactly the same issue and suggest adding a toggle "Enable notifications" just below the "Display events" in the calendar options.

ploink:
I have looked at the source code how to implement this, but need some feedback.

Adding a switch in the calendar settings is easy, but implementing the functionality is tricky and can be done in multiple ways:

1. In the android calendar database there is a column [ALLOWED_REMINDERS](https://developer.android.com/reference/android/provider/CalendarContract.CalendarColumns#ALLOWED_REMINDERS) to do just what we need, but it is readonly for calendar apps and can only be written by sync adapters. Etar may read and honor this setting, but changing it depends very much on the backend. For a local calendar it is impossible to change.
2. One alternative is to store a separate per-calendar setting in the Etar settings database. But, looking at the source code, there seems to be no implementation for storing per-calendar settings in Etar?
3. Another approach is to implement per-calendar notification channels. For a user point of view this is the most flexible as one has full control over notification settings. 

Maybe (1) should be implemented anyway and then either (2) or (3)?
Any ideas?

runejuhl:
Thank you for looking into this @ploink!

> Another approach is to implement per-calendar notification channels. For a user point of view this is the most flexible as one has full control over notification settings.

IMO this sounds like the most sensible (and flexible) approach.

Do you know if there are other calendar apps (free or not) that are able to disable notifications per calendar, and if so, how do they do it?

shompoe:
Yes!   

Simple Calendar Pro.

This is not an option within the app but it already uses a separate notification channel as mentioned above. Works very well for me. After recieving the first reminder for a calendar, the channel appears in android settings to disable it.

runejuhl:
@shompoe thank you -- sounds like separate notification channels are the way to go then.

[...]

Issue #67, opened 2016-06-29
Turn alerts on / off per calendar


akki42:
I would like to turn alerts (for upcoming events) on or off on a per calendar setting (instead of by a general setting for all calendars).

Use case: multiple calendars for various family members; it's great to see everybody's schedule, but annoying to get alerts for each other's appointments.

Would that be possible? Many thanks & keep up the great work!

Gitsaibot:
I would also love this feature ;). Is there someone how can do it? I'm not familiar enough with the notifications in android right now.

madduck:
This would be great to have!

smichel17:
You can do this with Calendar Notifications (Advanced calendar notifications) - https://f-droid.org/app/com.github.quarck.calnotify

2ip1:
The repository https://github.com/quarck/CalendarNotification has been archived and is no longer maintained. It would be nice to see this feature in Etar.

mueller-ma:
From #783 (seems to be a duplicate):

> Another approach is to implement per-calendar notification channels. For a user point of view this is the most flexible as one has full control over notification settings.

This is probably the easiest way to implement as all UI is already implemented by the OS.

PR #1571, opened 2024-01-22
Split calendar notifications into channels

### LINKED ISSUE #1493: Distinct notifications between calendars
Hi

Could you allow distinct notification chanels between calendars so we can set different notification sounds.
In my case, my wife shares her calendar with me, and so I would like to turn notifications off for it. With distinct chanels Android will allow tha out of the box.

Thanks

Wv5twkFEKh54vo4tta9yu7dHa3:
Closes #1493, closes https://github.com/Etar-Group/Etar-Calendar/issues/67, closes https://github.com/Etar-Group/Etar-Calendar/issues/783, closes https://github.com/Etar-Group/Etar-Calendar/issues/1138

### DIFF
--- a/app/src/main/java/com/android/calendar/AllInOneActivity.java
+++ b/app/src/main/java/com/android/calendar/AllInOneActivity.java
@@ -267,11 +267,11 @@ public class AllInOneActivity extends AbstractCalendarActivity implements EventH
         // This needs to be created before setContentView
         mController = CalendarController.getInstance(this);
 
-        // Create notification channel
-        AlertService.createChannels(this);
-
         // Check and ask for most needed permissions
         checkAppPermissions();
+
+        // Create notification channels
+        AlertService.createChannels(this);
 
         // Get time from intent or icicle
         long timeMillis = -1;
--- a/app/src/main/java/com/android/calendar/settings/CalendarPreferences.kt
+++ b/app/src/main/java/com/android/calendar/settings/CalendarPreferences.kt
@@ -25,6 +25,7 @@ import android.content.pm.PackageManager
 import android.graphics.drawable.Drawable
 import android.os.Bundle
 import android.provider.CalendarContract
+import android.provider.Settings
 import android.util.TypedValue
 import androidx.appcompat.app.AlertDialog
 import androidx.core.content.ContextCompat
@@ -34,6 +35,7 @@ import androidx.preference.PreferenceCategory
 import androidx.preference.PreferenceFragmentCompat
 import androidx.preference.SwitchPreference
 import com.android.calendar.Utils
+import com.android.calendar.alerts.channelId
 import com.android.calendar.persistence.CalendarRepository
 import ws.xsoh.etar.R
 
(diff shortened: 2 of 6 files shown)


app/src/main/java/com/android/calendar/AllInOneActivity.java
```diff
@@ -267,12 +267,12 @@ protected void onCreate(Bundle icicle) {
...
         mController = CalendarController.getInstance(this);
 
-        // Create notification channel
-        AlertService.createChannels(this);
-
         // Check and ask for most needed permissions
         checkAppPermissions();
 
+        // Create notification channels
+        AlertService.createChannels(this);
```
Wv5twkFEKh54vo4tta9yu7dHa3, line 274: Moving this after the permissions check, because the notification channel creation now needs calendar permission or it doesn't run

Wv5twkFEKh54vo4tta9yu7dHa3:
That code is only called by the `createChannels` function which has a `if (Utils.isOreoOrLater())` guard around its content (because channels only exist for Oreo and later, so function calls regarding channels need Oreo or later)

Wv5twkFEKh54vo4tta9yu7dHa3:
It's not possible to change it in the app, but I'll try to add a link to the channel setting

Wv5twkFEKh54vo4tta9yu7dHa3:
How's this?
image

Wv5twkFEKh54vo4tta9yu7dHa3:
Any other things you need to get this merged? :)

app/src/main/java/com/android/calendar/settings/CalendarPreferences.kt
```diff
@@ -132,12 +134,25 @@ class CalendarPreferences : PreferenceFragmentCompat() {
             title = getString(R.string.preferences_list_add_offline_message)
             isSelectable = false
         }
-
+fj
```
Gitsaibot, line 137: What is this ?

app/src/main/java/com/android/calendar/settings/CalendarPreferences.kt
```diff
@@ -132,12 +134,25 @@ class CalendarPreferences : PreferenceFragmentCompat() {
...
 
         if (!isLocalAccount) {
             screen.addPreference(synchronizePreference)
         }
+
         screen.addPreference(visiblePreference)
+
+        if(Utils.isOreoOrLater()){
+            val notificationPreference = Preference(context).apply {
+                title = getString(R.string.manage_notifications)
```
Gitsaibot, line 147: Unresolved reference: manage_notifications

app/src/main/java/com/android/calendar/settings/CalendarPreferences.kt
Wv5twkFEKh54vo4tta9yu7dHa3: Bonus letters I guess, sorry :see_no_evil:

app/src/main/java/com/android/calendar/settings/CalendarPreferences.kt
Wv5twkFEKh54vo4tta9yu7dHa3: Oops I changed the string to conform to the existing ones, but didn't change it in the code :facepalm:

Gitsaibot (approved):

jondo:
(Note: F-Droid rollout is blocked by https://github.com/Etar-Group/Etar-Calendar/issues/1607.)

Zocker1999NET:
Really, Thanks for implementing this feature, I'm sure this will also be helpful for me in the future.

(I do not know yet a better place to post this, so I post it here.)
However I want to add one thing: For me this was a breaking change because I configured the former notification channel to be able to override "Do not disturb" so I can use "Do not disturb" without blocking calendar notifications (Android has no setting for this for the whole app).
After this update, I missed important reminder notifications because the new notifications were not able to override Do not Disturb. It took me a week until I realized that this was the issue.
So please, the next time, announce this as a breaking change (with a pop-up in the app, or by making this feature optional, or at least by marking it in the release notes as a breaking change).

Wv5twkFEKh54vo4tta9yu7dHa3:
@Zocker1999NET Sorry I really did not think of this usecase. I don't think there is much we should do about it now, two months later, but I hope the impact was not too great for you!

PR #1603, opened 2024-03-07
Extend the distance between swiping left and right to better distingu…

### LINKED ISSUE #1535: scrolling down scrolls back up
When I try to scroll down, moving my thumb over the screen towards it's top and then releasing my thumb in the upper third, the content that was directly under my thumb when releasing just scrolls back down, even further, than from where i dragged it up.

I have this problem with versions 1.0.37/38. Not .36 from fdroid or .35 from play store.

I am using a Pixel 7 phone with Android 14.

Gitsaibot:
…ish between scrolling up or down. Fix #1535


toolstack:
Looks fine to me, 300 work well in my testing.  Perhaps a better/alternate solution might be to track the y motion as well and check if the swipe x/y ratio is more up/down than left/right.

toolstack:
Let me take a look at it, should be simple enough to implement.

toolstack:
PR #1605 implements the ratio check.

PR #1635, opened 2024-04-12
Convert Gradle Groovy Scripts to Kotlin Script

Doomsdayrs:
Instead of having to know two languages, only learn one.

Simplifies code base complexity and lowers barrier to entry for development.

All functionality preserved.


build.gradle.kts
Gitsaibot, line 1: Line 49 is needed as it is because of this: https://github.com/Etar-Group/Etar-Calendar/commit/66ef764245ebf6617d8b91715f2b2981e3ea47b2

Doomsdayrs: That's why I didn't touch it~

Doomsdayrs:
> @jspricke The more important question for me is whether the F-Droid build system can cope with this?

https://gitlab.com/fdroid/fdroiddata/-/blob/master/metadata/app.shosetsu.android.fdroid.yml

I develop Shosetsu, It uses KTS as well. F-Droid needed no changes from my memory, feel free to compare contrast the yamls.

theimpulson:
> @mikeNG Does the switch to "kts" work for your build process?

Replying on behalf of Mike, this is fine for us too

jspricke:
We need to find a better fix for #1641 before we can consider this.

Doomsdayrs:
> We need to find a better fix for #1641 before we can consider this.

Can you elaborate what needs to be done here? It looks like just updates to some dependence to me?

jspricke:
> Can you elaborate what needs to be done here? It looks like just updates to some dependence to me?

It revert the gradle syntax which produced random concurrency build failures. I think we should merge #1641 and see if the build stabilizes again for some releases before we do bigger work here.

jspricke:
> The thing is that the minor changes I made only affect the functions marked as deprecated by "lint". Since I don't know what the F-Droid build servers do and the build works with android studio I won't be able to do anything here. But the 1to1 conversion to kts scripts should work I think.

The F-Droid build server runs gralde from the command line on a fresh VM (no cache..), nothing special otherwise.

Doomsdayrs:
Rebased onto f1ac1e2c3d260d8f516d2a0cc5d8da8c9c7a3624, incorporating all changes.

This remains a 1-1 of the groovy code.

Doomsdayrs:
Huzzah, Check passes!

jspricke:
@Gitsaibot I would prefer a new release before that to get the latest version into F-Droid (I don't always have enough time to debug things like #1641).

Doomsdayrs:
Quick version update done~

Gitsaibot (approved):

PR #1641, opened 2024-04-14
Revert "Update gradle and fix warnings"

jspricke:
This reverts commit b8c2461684234163fa996a11878f3b13bcf66aa7.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,7 +24,7 @@ configure(external_projects) {
     }
 }
 
-tasks.register('aarGen') {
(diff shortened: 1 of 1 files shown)


jspricke:
This should fix the `java.util.ConcurrentModificationException` currently blocking the F-Droid builds.

Gitsaibot (approved):

jspricke:
I guess there is some implicit dependency modeled in the old syntax that is missing in the new.

Issue #1636, opened 2024-04-12
Compose Widget


Doomsdayrs:
I am leaving this issue to document my process on overhauling the current widget.

## Documentation
1. https://developer.android.com/develop/ui/compose/glance
    - It looks good. It's also an official guidance.

## Notes

- No architecture present in the application, with little separation of UI, State, and Business code.
  - This makes it troublesome to reuse the code from the current widget. And any other code in the application.
  - Possible separate issue to tackle for long term adaptability. 
- Will write new widget stack independent of current stack. Likely borrowing most code.

Doomsdayrs:
@jspricke @Gitsaibot since you both are people who I have seen active in this repo.

Which of you would be better to talk with about architectural problems of this application?

jspricke:
@Doomsdayrs The code is originally from the Google AOSP project and we only try to keep it alive. We welcome fixes for architectural problems as long as they are small logical self contained patches. We neither have the knowledge nor manpower to give you guidance for that.

Doomsdayrs:
> @Doomsdayrs The code is originally from the Google AOSP project and we only try to keep it alive. 

It's like if google was *trying* to ensure their code was unmaintainable. (sarcasm)

> We welcome fixes for architectural problems as long as they are small logical self contained patches.

A long series of small patches, I have experience with that. It's annoying compared to large overhauls, but I can get things done with enough free time. (Don't have currently)

> We neither have the knowledge nor manpower to give you guidance for that.

Wait, None of y'all have experience with system analysis and architecture?

jspricke:
> > We neither have the knowledge nor manpower to give you guidance for that.
> 
> Wait, None of y'all have experience with system analysis and architecture?

Not sure how you got to that conclusion but all I said is that we don't have the time help you.

Gitsaibot:
As far as the widget topic in general is concerned, I am "painless" when it comes to changes. I have never used the widget before and can therefore say little about the structure and architecture. That the code is not really great should be clear to anyone who has looked at it ;).

Doomsdayrs:
> > > We neither have the knowledge nor manpower to give you guidance for that.
> > 
> > 
> > Wait, None of y'all have experience with system analysis and architecture?
> 
> Not sure how you got to that conclusion but all I said is that we don't have the time help you.

I arrived at that question, not conclusion, due to the statement "neither have the knowledge nor manpower" which results in: `!haveKnowledge && !haveManpower`.

[...]

PR #1642, opened 2024-04-14
Split CalendarRepository into Repository & DataSources scheme

Doomsdayrs:
## Description

CalendarRepository is currently a monolithic file that
 contains all operations related to Calendars.
 While it is true that the Repository is meant to be a source of truth.
 It is not the case to have all operations in it.

Instead, splitting the repositorys data accessor functions into
 datasources lowers the code complexity of the repository itself,
 letting it focus on proper tasks such as actually handling dataflow
 and truth.

A final adjustment is adding an interface between the actual
 implementation of the repository and the UI layer.
 This allows UI developers to not be overwhelmed looking at
 implementation code, and instead focus on accessing methods they need.

Another small change made in this is replacing the usage of LiveData
 in the data layer.
 LiveData is a UI level structure meant best to be used with old
 android views, not for the data layer.

Replacement with Kotlin Flow fits the proper scheme for
 what the data layer is tasked with.
 To lower the amount of changes, the UI layer handles the change of the
 Flow into LiveData via the `asLiveData()` extension function.
 This change also removes ContentProviderLiveData.kt, which is simply
 replaced with `callbackFlow`.

## Testing

Tested by opening the application, adding in a new local calendar, changing colors.
Are there any other tests?

## Related

#1636


Doomsdayrs:
> Needs rebase.

Final weeks, been busy. Hopefully this handles it.

Gitsaibot (approved):

PR #1643, opened 2024-04-14
Add GPLv3 copyright profile

Doomsdayrs:
To ensure newly created files have copyright properly applied.

Is the license under "The Etar Project", since that's what the readme had the project referred to as, Or "Etar Group"?


jspricke:
As this is for specific editors(?) only, I would prefer to keep it out of the repository.

Doomsdayrs:
> As this is for specific editors(?) only, I would prefer to keep it out of the repository.

This is for Android studio. What other editor are you using to developer Android applications?

jspricke:
> This is for Android studio. What other editor are you using to develop Android applications?

vim.

Doomsdayrs:
> > This is for Android studio. What other editor are you using to develop Android applications?
> 
> vim.

That's a fascinating edge platform! I do wonder how your development flow goes.

---

This PR is important for those using the official recommend Android IDE, as it ensures that newly written code has the proper license header applied.

This ensures that new developers do not need to worry about headers, outside of changing an Apache header into a GPLv3 one without sufficient rewrite.

jspricke:
> This ensures that new developers do not need to worry about headers, outside of changing an Apache header into a GPLv3 one without sufficient rewrite (which only occurs if someone explicitly clicks "update copyright" for a file or all files, but its their problem at that point).

There is no need to change the header to GPLv3, unless you prefer to nor is there a need to sign away the author copyright.

Doomsdayrs:
> There is no need to change the header to GPLv3, unless you prefer to nor is there a need to sign away the author copyright.

I am confused by your comment, I am not promoting such action?

jspricke:
I don't use Android Studio but from reading this PR I would assume that it would:

> To ensure newly created files have copyright properly applied.

I assume this would default to GPLv3. Note that most of the Etar source code is Apache-2.0 licensed and we don't plan to change that. So I don't think we should enforce it in the Android Studio config.

> Is the license under "The Etar Project", since that's what the readme had the project referred to as, Or "Etar Group"?

Etar does not have a CLA so usually a contributor keeps their copyright by writing their name in the copyright header so again I would prefer if contributors would be able to change the Android Studio settings to their needs.

Issue #1666, opened 2024-05-26
goto day goes to wrong day


squintarelli:
I'm using the app on a Fairphone4 with /e/os, in Italian

while in the day view I select go to june 14th, I end up on June 28th:
I never end up to the desired day. 
it always misses by 6 to 15 days

Issue #1596, opened 2024-02-29
Incorrect date when using "Go to..."


Markoise:
Every time I want to jump to a certain date using "Go to...", a completely different day is shown to me.

E.g.:
- Select March 1st and jump to March 5th.
- Select June 1st and jump to June 28th

I'm using version 1.0.41 from the F-Droid Store on Android 14 on a Pixel 7.

Markoise:
If you only select one day in advance, then it works. But with several days in advance, you always jump way over the selected date.

Markoise:
The problem occurs in the day and week view. In the month and appointment overview, it works.

jpschewe:
I see something similar. For me every date that I selected takes me to the correct day of the week, but always in the last week of the month.

Markoise:
> I see something similar. For me every date that I selected takes me to the correct day of the week, but always in the last week of the month.

Unfortunately, I fear that this great app will no longer be developed. In any case, it's very quiet here.

Gitsaibot:
Please do not spread lies - this is not Russia!

Markoise:
Nobody wants to spread lies here.
I just hope that this great calendar will be developed further.

jspricke:
@Markoise please stop posting off topic messages, this is an issue tracker and not a discussion forum. Also this is in active development.

Issue #1251, opened 2022-10-22
Go to... goes to an incorrect week


Yenya:
When I use `Go to ...` from the three dots menu, it often goes to a different date than what I selected. E.g. now I have Etar in the week view, select November 25, and the week starting at November 14 ending at November 20 gets displayed.

I think the offset is usually -1 week, but I have seen different offsets as well as the correct week.

Etar 1.0.30 from F-Droid.

tribin:
Using "Go to..." from the three dots menu, each time I select a certain day, regardless if in the past or the future, Etar jumps to the day exactly >> four weeks after << the one I selected.
Using Etar 1.0.34 from F-Droid 1.16.3 on Android 8.0.0.
Anyway, thanks a lot for this fine and very useful App!

hartwork:
Same with Etar 1.0.34.

@etar-group please get in touch via my profile e-mail if there is interest in financial support for a timely bugfix and bugfix release for this issue.

jrcarter:
I have the Etar that comes with /e/ OS 1.16-S (Calendar "Version 12"). When I select a date in the future from "Go to ..." it takes me to the first date in the future on the same day of the week as the date I entered. For example, if today, Tuesday 2023-11-07, I say go to Thursday 2023-11-16, it takes me to Thursday 2023-11-09.

hartwork:
With Etar 1.0.36 now it seems to always jump to the current week, no matter what day I pick in the Go-To dialog.

I can only repeat sponsoring offer from earlier:

> Please get in touch via my profile e-mail if there is interest in financial support for a timely bugfix and bugfix release for this issue.

CC @Gitsaibot @jspricke

PR #1674, opened 2024-06-10
fix go-to date

jonas-haeusler:
setting the week-day here was once necessary when Etar was still using `android.text.format.Time`, where the days where 0-indexed (`java.util.Calendar`s days are 1-indexed).

doing this with `com.android.calendarcommon2.Time` is wrong, and will result in sundays being one week ahead.

additionally, setting the `DAY_OF_WEEK` without updating the `WEEK_OF_MONTH` will result in the old value of `WEEK_OF_MONTH` being used, as `Calendar` expects these to be set in combination when doing the computation: https://github.com/openjdk-mirror/jdk7u-jdk/blob/f4d80957e89a19a29bb9f9807d2a28351ed7f7df/src/share/classes/java/util/Calendar.java#L1729-L1740

fixes https://github.com/Etar-Group/Etar-Calendar/issues/1666
fixes https://github.com/Etar-Group/Etar-Calendar/issues/1596
fixes https://github.com/Etar-Group/Etar-Calendar/issues/1251

### DIFF
--- a/app/src/main/java/com/android/calendar/AllInOneActivity.java
+++ b/app/src/main/java/com/android/calendar/AllInOneActivity.java
@@ -935,15 +935,6 @@ public class AllInOneActivity extends AbstractCalendarActivity implements EventH
                     selectedTime.setMonth(monthOfYear);
                     selectedTime.setDay(dayOfMonth);
 
-                    Calendar c = Calendar.getInstance();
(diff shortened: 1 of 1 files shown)


Gitsaibot (approved):

hartwork:
@jonas-haeusler thank you! :+1:

PR #1719, opened 2024-08-14
Match `colorPrimaryDark` with `app_bar_color`

alexkoala:
This will make the App Bar look more unified with the Status Bar.


jspricke:
Thanks for your pull request, can you add a screenshot from before and after?

PR #1770, opened 2024-10-08
start event activity from notification instead trampolining via service

### LINKED ISSUE #1614: Clicking on reminder notification does not bring up event
I'm using Etar 1.0.41 on grapheneos with a calendar synced with davx.  When I get a notification for an upcoming event, clicking on it does not bring up the event detail.  I think this used to work fine, but I'm not sure what changed in the mean time.

Let me know what other debugging info would be useful.

jonas-haeusler:
as explained in https://github.com/Etar-Group/Etar-Calendar/issues/1614#issuecomment-2400537463

fixes #1614


app/src/main/java/com/android/calendar/alerts/DismissAlarmsService.java
```diff
@@ -114,15 +109,6 @@ public void onHandleIntent(Intent intent) {
                     (NotificationManager) getSystemService(Context.NOTIFICATION_SERVICE);
```
Gitsaibot, line 109: Do we need this anymore if we use `notificationBuilder.setAutoCancel(true);`

jonas-haeusler: yes,`DismissAlarmService` is still used for other notification actions, where `setAutoCancel` does not take effect

app/src/main/java/com/android/calendar/alerts/AlertReceiver.java
```diff
@@ -163,9 +161,12 @@ public static void finishStartingService(Service service, int startId) {
     }
 
     private static PendingIntent createClickEventIntent(Context context, long eventId,
-            long startMillis, long endMillis, int notificationId) {
-        return createDismissAlarmsIntent(context, eventId, startMillis, endMillis, notificationId,
-                DismissAlarmsService.SHOW_ACTION);
+        long startMillis, long endMillis) {
+        Intent intent = AlertUtils.buildEventViewIntent(context, eventId, startMillis, endMillis);
```
jonas-haeusler, line 165: todo: after receiving the intent in `EventInfoActivity` we have to update the state of the alert from `STATE_FIRED` to `STATE_DISMISSED` to ensure that the notification will not be displayed again.

see https://developer.android.com/reference/android/provider/CalendarContract.CalendarAlertsColumns#STATE

app/src/main/java/com/android/calendar/alerts/DismissAlarmsService.java
jonas-haeusler: yes, `DismissAlarmsService` is still used for other notification actions where `setAutoCancel` does not take effect. in fact, `setAutoCancel` does only work when you click on the notification content, but not on any notification actions.

jonas-haeusler:
> While you are at it, can you please fix the version check in AlertUtils and DismissAlarmsService.

done (https://github.com/Etar-Group/Etar-Calendar/pull/1770/commits/6bf5c8e4b36fcdc0cc168281cecca4c838286a53).

 i also removed code related to the `GlobalDismissManager` (da2a890cb21a315f30d3f4b5d31ca87e7741981b), which appeared to be unused. let me know if this is too much for this pull request, then i can drop that commit.

otherwise, this is done. please check whether this is working for you.

jspricke:
I am fine with this code wise but did not test it.

Gitsaibot (approved):

PR #1845, opened 2025-03-29
Fix event creation from Month view sometimes default to the wrong day

### LINKED ISSUE #1531: Suggestion: Less surprising event creation for empty days in Month view
For context: In my calendar, most days have no entries and days that aren't usually only have one or two entries (4 entries on a single day is very, very rare for me). As a result I pretty much exclusively use Month view. When not using Month view, I'd like to use Agenda view.

When in month view, if I tap on a day it goes to the Agenda view for that day (or testing now, I see that if I last used Day view then it goes to Day view instead which might actually be helpful for me, but let's set that aside for now). If the selected day doesn't have any events then Agenda view doesn't really have anything to show, so it "goes to" a different day meaning that if I follow up by tapping the "+" to add an event, it defaults to another day, seemingly whatever the nearest day was that had an event. In hindsight, as a developer I am unsurprised that the implementation ends up working this way. However as a user I find this very strange and confusing (and I cannot imaging that this is a specifically-intended behaviour).

A quick search through other issues reveals that I am not the only one. Issues #1147, #1148 both seem to be for this exact issue, and possibly #1023, #940 and #77.

I know you can long-press on a day to add an event but:
1. That's not intuitive and there's no way of knowing that this is an option. I didn't know you could do this until I came looking for related issues here on GitHub!
2. It doesn't give you the full "New event" form, you can only set a title and you always get an all-day event, which means that very often I'd need to separately go back to edit the event after creation.
3. In much the same way as extra clicks, long-tap is more work/effort
4. It's inconsistent with other views. On "Day" and "Week" view, if I tap an empty slot it turns blue and shows a "New event" button, which I can tap to get to the usual "Create event" screen. This means I can quickly double-tap to create an event.

So my suggestion has two options/forms:
1. Make month view consistent with the Day and Week views. Tapping on a day that has events in it goes to Agenda/Day view (depending on what you opened last, just like it does right now), but tapping on a day that does *not* have events turns the day into a "New event" button which can be tapped again to open the full "Create event" form. This makes the behaviour consistent across views (reducing surprise for users), and makes creating events from Month view significantly less tedious by allowing the form to be pre-filled with the correct date.
2. Alternatively (or possibly as an extension of the above): Make the behaviour consistent across Day, Week and Month view, but remove the 2nd tap to create an event. This would reduce the number of clicks required to create an event with the correct initial date & time from 2 to just 1 and honestly I'm not sure what else I could possibly want when tapping on an empty slot in my calendar if not to create an event. The only justification that I can think of for requiring the extra tap is to avoid navigating away from the current view if you accidentally tap somewhere once. I would expect that to be a pretty rare occurrence though, and the solution is just to hit the back button (or just adjust the datetime if you tapped on the wrong slot), making it 2 taps for the accidental flow and 1 tap for the "correct"/no-mistakes flow (instead of the other way around, which is what it currently is).

Regardless, thank you for all your work on Etar! I've been using it for years now and on the whole I find it excellent for my needs :)

jacquesh:
Previously, tapping on a day in month view would always take you to the agenda/detail view for that day. However if your calendar contains more events than will fit into the available vertical space (which most calendars do), it'll switch to agenda view and then immediately process an "on scroll" callback, which will cause the "selected date/time" to update as if the user had scrolled through their agenda.

Since this update changes the selected timestamp to whatever is currently visibly shown on the agenda, the scroll will always change the selected date to another day that *does* have an existing event.

The result of all of this is that if you have a full calendar with some empty days, and you tap on those days and then tap the "add event" widget, you probably won't check the date (I often don't) so it'll create an event on the wrong date and confuse you later.

This commit works around that problem by changing the behaviour of month view: If you tap on a day that already has events then it'll keep the existing behaviour of changing to the detail view. However if you tap on a day that has no events at all, it will go straight to the "create an event" view.

Personally, I would anticipate that creating an event is what users are trying to do most of the time when they tap on an empty day anyway, so in theory this also saves an extra tap for a common workflow.

This resolves #1531.

### DIFF
--- a/app/src/main/java/com/android/calendar/month/MonthByWeekAdapter.java
+++ b/app/src/main/java/com/android/calendar/month/MonthByWeekAdapter.java
@@ -328,10 +328,10 @@ public class MonthByWeekAdapter extends SimpleWeeksAdapter {
             // Else, check if the day we tapped has any events scheduled to it.
             int viewJulianDay = Time.getJulianDay(day.normalize(), day.getGmtOffset());
             int dayIndex = viewJulianDay - mFirstJulianDay;
-            boolean dayHasEvents = mEventDayList.subList(dayIndex, dayIndex+1)
+            boolean dayHasEvents = mEventDayList.subList(dayIndex, dayIndex + 1)
                      .stream().anyMatch(l -> !l.isEmpty());
 
-            if(dayHasEvents) {
+            if (dayHasEvents) {
                 // If there are events on that day, switch to the detailed view for that day
                 mController.sendEvent(mContext, EventType.GO_TO, day, day, -1,
                         ViewType.DETAIL,
--- a/app/src/main/java/com/android/calendar/month/MonthByWeekAdapter.java
+++ b/app/src/main/java/com/android/calendar/month/MonthByWeekAdapter.java
@@ -339,9 +339,8 @@ public class MonthByWeekAdapter extends SimpleWeeksAdapter {
                                 | CalendarController.EXTRA_GOTO_BACK_TO_PREVIOUS, null, null);
             } else {
                 // If there are *no* events on that day, let the user create one
-                long selectedTimeInMillis = mSelectedDay.normalize();
                 mController.sendEventRelatedEventWithExtra(this, EventType.CREATE_EVENT, -1,
-                        selectedTimeInMillis, 0, 0, 0,0, -1);
+                        day.normalize(), 0, 0, 0,0, -1);
             }
         }
     }
--- a/app/src/main/java/com/android/calendar/month/MonthByWeekAdapter.java
+++ b/app/src/main/java/com/android/calendar/month/MonthByWeekAdapter.java
@@ -325,11 +325,24 @@ public class MonthByWeekAdapter extends SimpleWeeksAdapter {
             mController.sendEvent(mContext, EventType.GO_TO, day, day, -1,
                     ViewType.CURRENT, CalendarController.EXTRA_GOTO_DATE, null, null);
         } else {
-            // Else , switch to the detailed view
-            mController.sendEvent(mContext, EventType.GO_TO, day, day, -1,
-                    ViewType.DETAIL,
-                            CalendarController.EXTRA_GOTO_DATE
-                            | CalendarController.EXTRA_GOTO_BACK_TO_PREVIOUS, null, null);
+            // Else, check if the day we tapped has any events scheduled to it.
+            int viewJulianDay = Time.getJulianDay(day.normalize(), day.getGmtOffset());
+            int dayIndex = viewJulianDay - mFirstJulianDay;
+            boolean dayHasEvents = mEventDayList.subList(dayIndex, dayIndex+1)
+                     .stream().anyMatch(l -> !l.isEmpty());
+
+            if(dayHasEvents) {
+                // If there are events on that day, switch to the detailed view for that day
+                mController.sendEvent(mContext, EventType.GO_TO, day, day, -1,
+                        ViewType.DETAIL,
+                        CalendarController.EXTRA_GOTO_DATE
+                                | CalendarController.EXTRA_GOTO_BACK_TO_PREVIOUS, null, null);
+            } else {
+                // If there are *no* events on that day, let the user create one
+                long selectedTimeInMillis = mSelectedDay.normalize();
+                mController.sendEventRelatedEventWithExtra(this, EventType.CREATE_EVENT, -1,
+                        selectedTimeInMillis, 0, 0, 0,0, -1);
+            }
         }
     }
 


jacquesh:
> For me, it is always the day view. There is one exception: when I switch from the agenda view to the monthly view and then tap on a day, I always end up back in the agenda view. This can be “fixed” by switching to the day view and then to the month view or close etar completely and open it again.

Yes, this is why I say "detailed view", because that's what it's called in the code and it goes to whatever the view is that you last used. I consider this a feature because I would rather have it go to agenda view than day view when I tap on a day from the month view. The fact that it remembers my last view means I can have that at the same time that you can have it go to day view :) Fully closing Etar does not change this setting for me though, I basically never even see Day view regardless of what I do with the app >.>

> I can't reproduce that in my case. For me, Etar always switches to the day view regardless of the number of appointments on that day. If the vertical space is not sufficient, I see a “+2” and so on.

Ah, I could have phrased this better. Let me put it differently: I specifically use agenda view, so for me when I tap on a day from month view I get the agenda. That's true always no matter the contents of the calendar. If I have an empty calendar with no events in it (e.g if I create a new virtual device or wipe an existing one so that it's completely fresh when launched from Android Studio), then it works as intended for a bit (IE if you add an event from agenda view, it auto-populates the correct date), which is why I mention having enough events to fill the space. If you have an empty calendar it won't reproduce the behaviour. This makes sense because if everything fits into the available space then there's no need to scroll, so the scroll callback doesn't run when you switch to that view, so it never updates the "currently selected date-time" to be another day that has calendar events. So if you're trying to reproduce the issue on a fresh virtual device, you need to:
1. Add enough events that they can't all be visible simultaneously in agenda view
2. Switch to agenda view if you haven't already so that it sets that as the detail view
3. Go back to month view and tap on a day that doesn't have any events on it.

So for example, here's my Pixel 3a AVD (which was freshly wiped and then I just ran Etar, added a calendar and the events visible below):
image
If I tap on the 27th or 28th, I get this:
image
and then if I tap the "+" in the bottom-right to add an event I get this:
image
which you'll notice has the wrong date.

> This can also be done by pressing and holding. But you save one tap ;)

I am aware, and I did call this out in the issue I linked above but there are several drawbacks to it that I mentioned there. I'll quote them below for convenience:
> 1. That's not intuitive and there's no way of knowing that this is an option. I didn't know you could do this until I came looking for related issues here on GitHub!
> 2. It doesn't give you the full "New event" form, you can only set a title and you always get an all-day event, which means that very often I'd need to separately go back to edit the event after creation.
> 3. In much the same way as extra clicks, long-tap is more work/effort
> 4. It's inconsistent with other views. On "Day" and "Week" view, if I tap an empty slot it turns blue and shows a "New event" button, which I can tap to get to the usual "Create event" screen. This means I can quickly double-tap to create an event.

app/src/main/java/com/android/calendar/month/MonthByWeekAdapter.java
```diff
@@ -325,11 +325,24 @@ protected void onDayTapped(Time day) {
...
-                    ViewType.DETAIL,
-                            CalendarController.EXTRA_GOTO_DATE
-                            | CalendarController.EXTRA_GOTO_BACK_TO_PREVIOUS, null, null);
+            // Else, check if the day we tapped has any events scheduled to it.
+            int viewJulianDay = Time.getJulianDay(day.normalize(), day.getGmtOffset());
+            int dayIndex = viewJulianDay - mFirstJulianDay;
+            boolean dayHasEvents = mEventDayList.subList(dayIndex, dayIndex+1)
+                     .stream().anyMatch(l -> !l.isEmpty());
+
+            if(dayHasEvents) {
```
Gitsaibot, line 334: Missing space: `if (dayHasEvents)`

app/src/main/java/com/android/calendar/month/MonthByWeekAdapter.java
```diff
@@ -325,11 +325,24 @@ protected void onDayTapped(Time day) {
...
         } else {
-            // Else , switch to the detailed view
-            mController.sendEvent(mContext, EventType.GO_TO, day, day, -1,
-                    ViewType.DETAIL,
-                            CalendarController.EXTRA_GOTO_DATE
-                            | CalendarController.EXTRA_GOTO_BACK_TO_PREVIOUS, null, null);
+            // Else, check if the day we tapped has any events scheduled to it.
+            int viewJulianDay = Time.getJulianDay(day.normalize(), day.getGmtOffset());
+            int dayIndex = viewJulianDay - mFirstJulianDay;
+            boolean dayHasEvents = mEventDayList.subList(dayIndex, dayIndex+1)
```
Gitsaibot, line 331: Missing space: `dayIndex + 1`

jacquesh: Fixed. I do have editorconfig support enabled so I'm not sure why android studio isn't detecting/fixing these but...¯\\\_(ツ)_/¯ Fixed regardless.

Gitsaibot (approved):

PR #1884, opened 2025-06-15
Integrate colorpicker sources

jspricke:


jspricke:
@theimpulson @chirayudesai would that work for you? (This is in preparation for #1859)

mikeNG:
Looks good to me, we'll figure out translations downstream and ideally we'll work on unifying downstream translations like we have discussed in the past.

Gitsaibot (approved):

PR #1885, opened 2025-06-23
Change timezonepicker to subtree

jspricke:


Gitsaibot (commented):
timezonepicker instead of colorpicker.

Issue #1906, opened 2025-07-19
Date picker is set to today when modifying an event


jspricke:
> If you edit an event that's on a different date than *today*, and then tap to edit the start date, the date in the picker will be set to *today*. So if you at this point just hit "Save", the date will be set to *yesterday* (because of the original bug reported here).
> 
> If this is legitimate and a separate bug, I can file it separately, if you prefer. 

 _Originally posted by @orblivion in [#1904](https://github.com/Etar-Group/Etar-Calendar/issues/1904#issuecomment-3091254085)_

orblivion:
Specifically, it seems, it's set to *UTC today*.

Issue #1904, opened 2025-07-18
Selecting a date for an event is one day off


CaviaPorcellus:
In the event creation/editing screen, if I select a day for an event, either a new event or editing an older one, the day before is selected.

To reproduce:
1.) Go to the event edit screen.
2.) Press the start date button.
3.) Select a day for the even to start.

Result:
The event day is one day before that which was selected (e.g., if I select July 23, July 22 is the event date).

I'm using Etar 1.0.49 from F-Droid on Android 16 (Google Pixel 8a). This did not happen in version 1.0.48.

July 22 is selected from the date picker:
<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/d470f90d-17bd-4980-9ebd-d6cf21708fdb" />

July 21 is instead set as the date:
<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/61510c3c-299a-48ec-87f1-8015d79be9cb" />

landry314:
omg this is so horrible i need to downgrade immediately but there is no backup system so now i am struggling. please please please fix this immediately.

and please release the apk on github because i cant wait 3 days for f-droid to release it. please.

Gitsaibot:
Interesting can't reproduce this on my phone 🤔 

@jspricke, @theimpulson Can you reproduce?

landry314:
Yes, that was it. I lowered the system font size by one and it was fixed.

BUT, it would be nice if the app adapted the space to the system font size. My vision has gotten worse and I need a higher font size.

landry314:
Maybe it is a time zone issue where it thinks it is a different day but for me i am just choosing the day after for every event I make or edit.

Graphene OS (Android 16) on Pixel 8

orblivion:
Ditto, was just planning on reporting this. F-Droid version, GrapheneOS (Android 16) Pixel 7. I think there were a couple cases where it didn't happen. I couldn't quite figure out those conditions. I wonder if those unable to reproduce hit on one of those. Maybe try a few times, creating new events on different dates, editing existing events, etc?

Another thing I noticed that might be related, might be a separate bug, or might be expected behavior that I just forgot about:

If you edit an event that's on a different date than *today*, and then tap to edit the start date, the date in the picker will be set to *today*. So if you at this point just hit "Save", the date will be set to *yesterday* (because of the original bug reported here).

If this is legitimate and a separate bug, I can file it separately, if you prefer.

[...]

PR #1909, opened 2025-07-19
fix datePicker returns incorrect date

Gitsaibot:
Fix for #1904 and #1906

### DIFF
--- a/app/src/main/java/com/android/calendar/event/EditEventView.java
+++ b/app/src/main/java/com/android/calendar/event/EditEventView.java
@@ -1664,7 +1664,7 @@ public class EditEventView implements View.OnClickListener, DialogInterface.OnCa
 
     private void showDatePickerDialog(View view) {
         MaterialPickerOnPositiveButtonClickListener<Long> materialPickerOnPositiveButtonClickListener = timePicked -> {
-            Calendar calendar = Calendar.getInstance();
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/android/calendar/event/EditEventView.java
```diff
@@ -1664,7 +1664,7 @@ public void onTimeSet(View view, int hourOfDay, int minute) {
 
     private void showDatePickerDialog(View view) {
         MaterialPickerOnPositiveButtonClickListener<Long> materialPickerOnPositiveButtonClickListener = timePicked -> {
-            Calendar calendar = Calendar.getInstance();
+            Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
```
theimpulson, line 1667: Should this be the local timezone? I don't know what the behaviour was earlier, but if it was the same, it should be okay.

theimpulson (commented):
Looks good to me other than a minor question

app/src/main/java/com/android/calendar/event/EditEventView.java
Gitsaibot: I don't understand your question. The Thing is that the getInstance() method in Calendar class is used to get a calendar using the current time zone and locale of the system. But time picked is a Timestamp in UTC as far as i understand. This will lead to the behaviour described in #1904

PR #1931, opened 2025-07-27
Improve guest selection style so that the colors match the different …

### LINKED ISSUE #1326: Dark theme: guest selection invisible
The guest selection in new events is very hard to read. Because of black text on dark grey background. 

Etar 1.0.29 from F-Droid on Android 13 (Samsung Galaxy Note20 5G/OuneUi 5.0)

Gitsaibot:
…themes. Also added support for vectorDrawables.

Fix #1326

### DIFF
--- a/app/src/main/java/com/android/calendar/chips/CircularImageView.java
+++ b/app/src/main/java/com/android/calendar/chips/CircularImageView.java
@@ -12,14 +12,15 @@ import android.graphics.Shader;
 import android.graphics.drawable.BitmapDrawable;
 import android.graphics.drawable.Drawable;
 import android.graphics.drawable.StateListDrawable;
+import android.graphics.drawable.VectorDrawable;
 import android.util.AttributeSet;
-import android.widget.ImageView;
+import androidx.appcompat.widget.AppCompatImageView;
 
 /**
  * An ImageView class with a circle mask so that all images are drawn in a
  * circle instead of a square.
  */
-public class CircularImageView extends ImageView {
+public class CircularImageView extends AppCompatImageView {
     private static float circularImageBorder = 1f;
 
     private final Matrix matrix;
@@ -58,12 +59,12 @@ public class CircularImageView extends ImageView {
     @Override
     protected void onDraw(Canvas canvas) {
         Drawable drawable = getDrawable();
-        BitmapDrawable bitmapDrawable = null;
+        BitmapDrawable bitmapDrawable;
         // support state list drawable by getting the current state
         if (drawable instanceof StateListDrawable) {
-            if (((StateListDrawable) drawable).getCurrent() != null) {
-                bitmapDrawable = (BitmapDrawable) drawable.getCurrent();
-            }
+            bitmapDrawable = (BitmapDrawable) drawable.getCurrent();
+        } else if (drawable instanceof VectorDrawable) {
+            bitmapDrawable = getBitmapDrawableFromVectorDrawable(drawable);
         } else {
             bitmapDrawable = (BitmapDrawable) drawable;
         }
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/android/calendar/chips/CircularImageView.java
```diff
@@ -58,14 +59,17 @@ public CircularImageView(Context context, AttributeSet attrs, int defStyle) {
...
     protected void onDraw(Canvas canvas) {
         Drawable drawable = getDrawable();
-        BitmapDrawable bitmapDrawable = null;
+        BitmapDrawable bitmapDrawable;
         // support state list drawable by getting the current state
         if (drawable instanceof StateListDrawable) {
-            if (((StateListDrawable) drawable).getCurrent() != null) {
-                bitmapDrawable = (BitmapDrawable) drawable.getCurrent();
-            }
+            drawable.getCurrent();
```
jspricke, line 65: why is this called here? get implies there is a return value but it is not used.

I don't understand this change.

app/src/main/res/values/colors.xml
```diff
@@ -183,6 +183,20 @@
     <color name="recurrence_picker_background">#fff2f2f2</color>
     <color name="recurrence_bubble_text_normal">#ff737373</color>
 
+    <!-- Chips specific colors -->
```
jspricke, line 186: Is there a reason to move this?

app/src/main/res/values/colors.xml
```diff
@@ -286,9 +300,8 @@
     <!-- timezone_GMT_background_color -->
     <color name="calendar_date_banner_background_black">#212121</color>
 
-    <!-- Overwrite external chips colors -->
+    <!-- Chips colors -->
```
jspricke, line 303: maybe move all chip colors to one place?

app/src/main/java/com/android/calendar/chips/CircularImageView.java
```diff
@@ -111,4 +115,19 @@ public void drawBitmapWithCircleOnCanvas(Bitmap bitmap, Canvas canvas,
         canvas.drawCircle(dest.centerX(), dest.centerY(),
                 dest.width() / 2f - circularImageBorder / 2, borderPaint);
     }
+
+    private BitmapDrawable getBitmapDrawableFromVectorDrawable(Drawable drawable) {
```
jspricke, line 119: why do we need this? Can we maybe switch everything to vecor graphics?

Gitsaibot: We need this because our icon is a vector image, but the contact images are bitmap images, as you would expect.

app/src/main/res/values/colors.xml
Gitsaibot: The general colors are at the top, followed by the two parts with dark and black themes colors.

app/src/main/res/values/colors.xml
Gitsaibot: We can't that is for theme.

app/src/main/java/com/android/calendar/chips/CircularImageView.java
```diff
@@ -58,14 +59,16 @@ public CircularImageView(Context context, AttributeSet attrs, int defStyle) {
...
-                bitmapDrawable = (BitmapDrawable) drawable.getCurrent();
-            }
+            bitmapDrawable = (BitmapDrawable) drawable.getCurrent();
         } else {
-            bitmapDrawable = (BitmapDrawable) drawable;
+            if (drawable instanceof VectorDrawable) {
+                bitmapDrawable = getBitmapDrawableFromVectorDrawable(drawable);
+            } else {
+                bitmapDrawable = (BitmapDrawable) drawable;
+            }
```
jspricke, line 71: move the this one level up, i.e.:
```
if (drawable instanceof StateListDrawable) {
} else if (drawable instanceof VectorDrawable) {
} else {
}
```

PR #1939, opened 2025-08-01
Bump targetSdk to 35

luk1337:


luk1337:
yeah, your pr works fine here too.

Issue #1801, opened 2024-11-20
Auto-dismissal at event end time means missed events never get noticed


nickleverton:
I sometime miss events that I have logged in my calendar (these are mostly less urgent or optional events so I don't set tons of reminders).  This is because I don't always have my phone on me, or may have it on silent or DND during some other task.

It has been useful that the notification for these events remains active until I notice and, when I am ready, I can dismiss it manually.  This means I can catchup, in whatever manner is needed, at a time of my choosing (download video stream from the event or apologise to the host etc etc).

I'm concerned that the change in commit 611b1a5, to auto-dismiss event notifications at end time, may mean that I can't rely on Etar for this function of reminding me that I have missed something. 

I admit that I haven't upgraded to 1.0.48 yet because of this concern but this appears to be the effect of this change, please forgive me if I have misunderstood.

I am not sure whether this change was made in response to some other issue, but would it be possible to revert it please ?  I might have asked for it to be an option but I recognise that too many fiddly little options can provide a confusing UI experience.

Thankyou either way for a brilliant piece of software in Etar, which I rely on daily and is far better than the proprietary calendars I have to use on work-issued devices.

jspricke:
Wouldn't a ToDo list serve your use case better?

nickleverton:
Hi, thankyou for your suggestion  I have todo list apps but they are for goals that need doing and must be kept track of in time.

The Etar model of creating an event is both a better conceptual match for these real events, and also suits my use much better for things that I may or may not attend.

ghgr:
I agree with @nickleverton. This is a reasonable use case (compared, to, say, [this one](https://xkcd.com/1172/)), and it might be better served with e.g. a flag. It's been a long time since I used JAVA and I'm not familiar with the codebase, so forgive me if I'm wrong, but looking at this [commit](https://github.com/Etar-Group/Etar-Calendar/commit/611b1a5bdb879757476978edb08480574ef2f1f9) could it be implemented with something like?

```java
if (endTime < currentTime && flags.AUTODISMISS_ON_COMPLETION) {
    newState = CalendarAlerts.STATE_DISMISSED;
}
```

dexic79:
I second the request for not deleting the reminders. I am using /e/OS and I can't turn off the notifications completely to use another calendar, so I would better rely on the calendar I am stuck with anyway. 

There are many reasons to use the calendar for tasks that may or have to happen on this particular day or to remind you of preparing something a couple of days ahead.

[...]

PR #1942, opened 2025-08-02
Make auto dismissal optional and disable it by default

### LINKED ISSUE #1891: No notification, if the reminder is 0 minutes
Hi,

If I check "After start", the notification does not work either.

This option is very important for me, because I use it for tasks

c0d3z3r0:
Closes: #1801

Also sort-of resolves #1891


jspricke:
@Gitsaibot I fully agree.

c0d3z3r0:
I'm fine with either way. I just had assumed just reverted would get contradiction, because the discussion in #1801 didn't make much progress :) Revert PR: #1944

PR #1954, opened 2025-08-22
Cleanup

Gitsaibot:


jspricke:
> The check doesn't actually change anything, it just prevents a crash(OutofBounds) if the column doesn't exist.

Right, but the `it.moveToFirst()` changes the pointer, is that no longer needed?

jspricke (approved):

PR #2017, opened 2025-12-31
Fix "black theme colors not loaded"

Gitsaibot:


app/src/main/java/com/android/calendar/theme/ThemeUtils.kt
```diff
@@ -27,13 +27,13 @@ object ThemeUtils {
...
             Theme.SYSTEM -> if (isSystemInDarkTheme(context)) {
-                return "_" + "dark"
+                return if (context.isPureBlackModeEnabled) "_" + "black" else "_" + "dark"
             } else {
                 return ""
             }
-
             Theme.LIGHT -> return ""
-            Theme.DARK, Theme.BLACK -> return "_" + theme.name.lowercase()
+            Theme.DARK -> return if (context.isPureBlackModeEnabled) "_" + "black" else "_" + "dark"
```
jspricke: The `if` looks a little odd given the extra `Theme.BLACK` case below, but I guess that's intended.

Gitsaibot: Yes it's intended but in future we should only offer one "dark" mode. It's less confusing

PR #2020, opened 2026-01-02
Add stacked events

### LINKED ISSUE #1976: [enhancement] Stacked view of overlapping events
The stock android calendar shows events happening at overlapping times in stacked order, which allows to still read what they are about, example:

Image

ETAR instead puts them side-by-side, which makes them unreadable (no visible text):

Image

rincewound:
This fixes #1976 by drawing events over each other similar to how Google Calendar does, allowing users to read the text of events that are happening in parallel:
image

### DIFF
--- a/app/src/main/java/com/android/calendar/DayView.java
+++ b/app/src/main/java/com/android/calendar/DayView.java
@@ -87,6 +87,7 @@ import com.google.android.material.dialog.MaterialAlertDialogBuilder;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Calendar;
+import java.util.Comparator;
 import java.util.Formatter;
 import java.util.Locale;
 import java.util.regex.Matcher;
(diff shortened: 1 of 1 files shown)


AntonArbandt:
Nice work :) just one thing:

If a event is in front of a earlier event it is hidden behind the next:
hidden

Clicking on the right of the purple event opens the menu for the event behind it.

Momro:
Just always put the new calendar entry on top of the previous one. I scrolled through a few weeks in my Google calendar and it seemed appropriate. It's also how Google does it

app/src/main/java/com/android/calendar/EventGeometry.java
```diff
@@ -97,8 +102,9 @@ public boolean computeEventRect(int date, int left, int top, int cellWidth, Even
             event.bottom = event.top + mMinEventHeight;
         }
 
-        float colWidth = (float) (cellWidth - (maxCols + 1) * mCellMargin) / (float) maxCols;
-        event.left = left + col * (colWidth + mCellMargin);
+
+        float colWidth = (float) (cellWidth - (maxCols) * mCellMargin) - (float) (col * mPerColMargin);
```
Gitsaibot, line 106: Can you please replace this with:

event.left = left + (col) * mPerColMargin;
event.right = left + cellWidth - mCellMargin;
return true;

This should fix the right border issue.

Issue #2019, opened 2026-01-01
Making a recurring event with Etar can cause errors with other caldav clients


bgstack15:
#### How to reproduce
* Use Etar app to create a recurring event, that is, one with a RRULE with UNTIL attribute.

```
RRULE:FREQ=WEEKLY;WKST=SU;UNTIL=20260428T130000;BYDAY=TU
```

#### What is expected
The caldav entry on the server will store UNTIL with a UTC timestamp as demonstrated in https://datatracker.ietf.org/doc/rfc5545/. This will allow other caldav clients to parse the rule.
```
RRULE:FREQ=WEEKLY;WKST=SU;UNTIL=20231116T191500Z;BYDAY=TH
```

#### What actually happens
The caldav entry on the server does not store the UNTIL in UTC, which breaks at least one other caldav client which complains about "Invalid UNTIL value: 20260428T130000"
Etar still works, though.

PR #2024, opened 2026-01-15
Time switchTimezone: Also update timezone attribute, Fixes recurrence end date

lennart-k:
When using `switchTimezone` the `timezone` attribute is not updated but only the internal Gregorian calendar.

In the case of the end date in the `RecurrencePickerDialog`, the chain of side effects
basically causes the `format2445` function to replace the `UTC` timezone with the wrong timezone.
I've documented what I think happened in https://github.com/Etar-Group/Etar-Calendar/issues/2019#issuecomment-3755107334.

This one line fixes the `RecurrencePickerDialog` generating invalid calendar data #2019.
I've also added a panic to ensure that we never generate a non-UTC end date.

### DIFF
--- a/app/src/main/java/com/android/calendar/calendarcommon2/Time.java
+++ b/app/src/main/java/com/android/calendar/calendarcommon2/Time.java
@@ -212,6 +212,7 @@ public class Time {
     }
 
     public void switchTimezone(String timezone) {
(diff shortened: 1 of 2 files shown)


Gitsaibot (approved):
Thanks for the PR and the good bug description. Looks good to me.

PR #2117, opened 2026-05-17
Allow any day of the week as the week start day

Zhouzhishi808:
## Summary
This PR extends the "Week starts on" preference to allow users to select **any day** of the week (Sunday through Saturday), instead of being limited to only Saturday, Sunday, or Monday.
## Motivation
Currently, Etar only offers four options for the week start day: Locale default, Saturday, Sunday, and Monday. While these cover many locales, some users have work schedules or payroll cycles that start on other days (e.g., Tuesday, Wednesday, Thursday, Friday). Allowing any day improves flexibility and usability.
## Changes
- **`GeneralPreferences.kt`**: Dynamically generates week start day labels from `DateFormatSymbols`, supporting all 7 days. Added constants for Tuesday–Friday (`WEEK_START_TUESDAY` through `WEEK_START_FRIDAY`).
- **`Utils.java`**: Replaced the hardcoded `if-else` chain in `getWeekStartDay()` with a generic `convertCalendarDayToTimeDay()` method that supports any day value.
- **`arrays.xml`**: Removed the hardcoded `preferences_week_start_day_labels` array. Expanded `preferences_week_start_day_values` to include values for all 7 days.
- **`general_preferences.xml` / `general_preferences.xml` (v26)**: Removed the static `app:entries` attribute, since labels are now set dynamically.
- **`strings.xml`**: Added `week_start_locale_default` string resource.
**Before**: Only Locale default, Saturday, Sunday, Monday  
**After**: Locale default + any day (Sunday, Monday, Tuesday, Wednesday, Thursday, Friday, Saturday)
## Testing
Built the debug APK, installed on a Pixel 7a (Android 16), verified that:
- All 7 days appear in the week start day dropdown
- Setting each day correctly updates the calendar view
- Default value (Locale default) still works correctly

### DIFF
--- a/app/src/main/java/com/android/calendar/Utils.java
+++ b/app/src/main/java/com/android/calendar/Utils.java
@@ -675,13 +675,11 @@ public class Utils {
             startDay = Integer.parseInt(pref);
         }
 
-        if (startDay == Calendar.SATURDAY) {
-            return Time.SATURDAY;
-        } else if (startDay == Calendar.MONDAY) {
-            return Time.MONDAY;
-        } else {
-            return Time.SUNDAY;
-        }
+        return convertCalendarDayToTimeDay(startDay);
+    }
+
+    private static int convertCalendarDayToTimeDay(int calendarDay) {
+        return calendarDay - 1;
     }
 
     /**
@@ -774,9 +772,7 @@ public class Utils {
      * @return true if the column is Saturday position
      */
     public static boolean isSaturday(int column, int firstDayOfWeek) {
-        return (firstDayOfWeek == Time.SUNDAY && column == 6)
-                || (firstDayOfWeek == Time.MONDAY && column == 5)
-                || (firstDayOfWeek == Time.SATURDAY && column == 0);
+        return column == (6 - firstDayOfWeek + 7) % 7;
     }
 
     /**
@@ -787,9 +783,7 @@ public class Utils {
      * @return true if the column is Sunday position
      */
     public static boolean isSunday(int column, int firstDayOfWeek) {
-        return (firstDayOfWeek == Time.SUNDAY && column == 0)
-                || (firstDayOfWeek == Time.MONDAY && column == 6)
-                || (firstDayOfWeek == Time.SATURDAY && column == 1);
+        return column == (-firstDayOfWeek + 7) % 7;
     }
 
     /**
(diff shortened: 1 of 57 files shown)


Zhouzhishi808:
Thanks for the review. I've addressed the hardcoded `firstDayOfWeek` logic in the latest commit.
**Changes:**
1. **`isSaturday()` / `isSunday()`** — replaced the `if-else` chain with generalized formulas using modulo arithmetic, so they work for all 7 possible `firstDayOfWeek` values.
2. **`getWeekNumberFromTime()`** — the original code only shifted dates for Sunday/Monday/Saturday week starts. Generalized to handle any non-Monday week start by computing the correct forward shift to the nearest ISO Monday.
3. **`SimpleDayPickerFragment.doResumeUpdates()`** and **`SimpleWeeksAdapter` constructor** — changed from `Calendar.getInstance(Locale.getDefault()).getFirstDayOfWeek()` to `Utils.getFirstDayOfWeek(context)` so they respect the user's preference instead of the system locale default.
Please take another look when you have a chance.

Zhouzhishi808:
**Cleaned up the diff:** the previous commit accidentally included CRLF→LF line ending changes across the files. I've redone everything on the original base files so the diff is now limited to the actual code changes only (8 files, +56/−42).

app/src/main/java/com/android/calendar/settings/GeneralPreferences.kt
```diff
@@ -526,6 +546,10 @@ class GeneralPreferences : PreferenceFragmentCompat(),
         const val WEEK_START_SATURDAY = "7"
```
Gitsaibot: Wrong order. See Text above ;)

Zhouzhishi808:
Fixed, reordered constants in ascending numerical order.

app/src/main/res/values/arrays.xml
```diff
@@ -125,18 +125,15 @@
         <item>1</item>
     </string-array>
 
-    <string-array name="preferences_week_start_day_labels">
```
Gitsaibot, line 128: This array must be removed from all other languages too

Zhouzhishi808:
Fixed the constant ordering and removed the preferences_week_start_day_labels array from all 49 locale arrays.xml files. Also rebased onto latest upstream.

app/src/main/java/com/android/calendar/Utils.java
```diff
@@ -1434,16 +1428,23 @@ public static int getWeekNumberFromTime(long millisSinceEpoch, Context context)
...
-        // isn't Monday we may need to shift the date to be in the correct
-        // week
-        if (weekTime.getWeekDay() == Time.SUNDAY
-                && (firstDayOfWeek == Time.SUNDAY || firstDayOfWeek == Time.SATURDAY)) {
-            weekTime.setDay(weekTime.getDay() + 1);
-            weekTime.normalize();
-        } else if (weekTime.getWeekDay() == Time.SATURDAY && firstDayOfWeek == Time.SATURDAY) {
-            weekTime.setDay(weekTime.getDay() + 2);
-            weekTime.normalize();
+        // Time.getWeekNumber() uses ISO 8601 (Monday-based). When the user's
```
Gitsaibot: According to AI, this section contains a logical error:

> The current implementation of getWeekNumberFromTime contains a logical error when the week starts on Tuesday, Wednesday, or Thursday.
> 
> The code uses a strategy of shifting days that fall between the "User's Start Day" and the "ISO Monday" forward to that Monday. It then calls Time.getWeekNumber(), which follows the ISO 8601 standard (where a week is strictly Monday to Sunday).
> The core of the issue is that in the ISO 8601 system, a week's number is defined by its Thursday. Any 7-day period (regardless of when it starts) should be labeled with the ISO week number of the Thursday it contains.
> By shifting days forward to the next Monday, the code often pushes the date into the next ISO week, causing the reported week number to be one week too high.
> 
> Example: Tuesday Start
> Imagine the user's week starts on Tuesday, Jan 2, 2024.
> 1. In the ISO system, Thursday, Jan 4 is the anchor for Week 1.
> 2. The user's week is: Tue (Jan 2), Wed (Jan 3), Thu (Jan 4), Fri (Jan 5), Sat (Jan 6), Sun (Jan 7), Mon (Jan 8).
> 3. Your current code takes Tuesday (Jan 2) and shifts it forward by 6 days to Monday, Jan 8.
> 4. Time.getWeekNumber() for Jan 8 returns Week 2.
> 5. Result: The calendar incorrectly displays "Week 2" for a week that should be "Week 1".
> 
> Instead of shifting to Monday, the logic should find the Thursday of the user's current week. Because every 7-day period contains exactly one Thursday, and that Thursday's ISO week number is the correct designation for that period, this is the most reliable anchor.
> 
> 1. Calculate how many days have passed since the start of the user's week: diff = (weekDay - firstDayOfWeek + 7) % 7.
> 2. Move the time to the start of the user's week: time - diff.
> 3. Move the time to the 4th day (Thursday) of that specific week: + 3.
> 4. Return the week number of that Thursday.
> 
> This ensures that regardless of whether the week starts on a Monday, a Saturday, or a Wednesday, it always returns the ISO-compliant week number for that specific block of 7 days.

Zhouzhishi808:
Changed getWeekNumberFromTime() to find the Thursday within the user's current week (ISO 8601 anchor day) instead of shifting to Monday. This correctly handles all week starts including Tuesday–Friday

Gitsaibot (approved):

Issue #1362, opened 2023-03-18
[feature request] layer overlapping items instead of shrinking their width (weekly view)


dudenotnamedben:
First of all thank you for the great work on the app.  

I use it with a shared calendar so it happens that items have overlapping timeframes/timeslots.

My feature request would be to display items that have an overlapping timeframe in a layered style. (Maybe distinguishing them with a border or a shadow.)
That way you could keep readability in a lot of cases where it's not possible now. 

Thanks again :).

Cwpute:
Ten times this.
When two events happen in the same time-frame, the shorter one should be on top.
Other calendars do that too, probably Google Calendars, but i'm not sure.

Cwpute:
I've provided a little preview to show what it could look like in Etar, in weekly view specifically, but of course it could be applied to other views.

1. This is what two overlapping events currently look like:
![current embedded](https://github.com/Etar-Group/Etar-Calendar/assets/80604512/5e4363dd-9f3d-41e9-a4dc-87dc9b07125f)
As you can see this doesn't give much information about the event, appart from its color.
2. This is what the two events look like separately. 
![embedded original events](https://github.com/Etar-Group/Etar-Calendar/assets/80604512/1a04b01f-8d14-4ed8-a490-c70087f19653)
3. This is what i'd like it to look like:
![embedded event](https://github.com/Etar-Group/Etar-Calendar/assets/80604512/418e8279-769c-4e33-9561-301a87804603)
The UI would need to be able to recognize words and cut it when it's close to a nested event, using `(…)` as a sign it's been cut off. Would be nice as well to detect available space to put any remaining information that couldn't be put before, as shown with the place of the event, here being "Ludothèque de Parthenay".

da-the-dev:
very much needed

danschwab:
I agree with dudenotnamedben, thank you for your great work and great app!

I also agree that overlapping events in week view are uninteligable. I would just show event title text nomatter how small events are. Remove the minimum size constraint. Make it an option and let users deal with the ui tradeoffs.

Cwpute:
> I would just show event title text nomatter how small events are.

This would definitely be a very good first step to make this more usable pretty quickly !

larsbonczek:
I have implemented this feature in my fork of this repo and am currently testing it in my daily life:

https://github.com/larsbonczek/Etar-Calendar/tree/staggered-display

[...]

PR #2119, opened 2026-05-27
Add option for staggered display of overlapping events

larsbonczek:
This implements the feature requested in #1362.

### DIFF
--- a/app/src/main/java/com/android/calendar/DayView.java
+++ b/app/src/main/java/com/android/calendar/DayView.java
@@ -25,6 +25,7 @@ import android.content.ContentResolver;
 import android.content.ContentUris;
 import android.content.Context;
 import android.content.DialogInterface;
+import android.content.SharedPreferences;
 import android.content.res.Resources;
 import android.content.res.TypedArray;
 import android.database.Cursor;
@@ -87,7 +88,9 @@ import com.google.android.material.dialog.MaterialAlertDialogBuilder;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Calendar;
+import java.util.Comparator;
 import java.util.Formatter;
+import java.util.List;
 import java.util.Locale;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -313,6 +316,7 @@ public class DayView extends View implements View.OnCreateContextMenuListener,
     private long mLastReloadMillis;
 
     private ArrayList<Event> mEvents = new ArrayList<Event>();
+    private ArrayList<Integer> mVisibleEventIndices = new ArrayList<Integer>();
     private ArrayList<Event> mAllDayEvents = new ArrayList<Event>();
     private StaticLayout[] mLayouts = null;
     private StaticLayout[] mAllDayLayouts = null;
@@ -429,6 +433,7 @@ public class DayView extends View implements View.OnCreateContextMenuListener,
     private static int EVENT_RECT_LEFT_MARGIN = 1;
     private static int EVENT_RECT_RIGHT_MARGIN = 0;
     private static int EVENT_RECT_STROKE_WIDTH = 2;
+    private static int STAGGERED_EVENT_SHADOW_WIDTH = 4;
     private static int EVENT_TEXT_TOP_MARGIN = 2;
     private static int EVENT_TEXT_BOTTOM_MARGIN = 2;
     private static int EVENT_TEXT_LEFT_MARGIN = 6;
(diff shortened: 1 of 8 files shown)


Cwpute:
i initially posted my feedback in the issue post but i figured i should have posted there instead. Here it comes :

Cwpute:
Image


It's great to see it in action !

Here is some feedback : 

1. I feel overlapping too much events can get messy like in @larsbonczek 's screenshot, especially when you put them next to other days full of events. The day this chain of overlapping events is a part of is not so clear anymore, as the clear column for each day is broken or twisted, and produces visual clutter. One way to solve that i think would be to have the longest event among a chain to use the full width available for that day, so that any shorter event actually appears like it's "notched" into it. The column-felling of each day would be more or less maintained i think. See the mock-up i made earlier in this issue, attached below.
Another solution would be to have the shorter events centered in the day-colum, instead of pushed to either side.
Image

2. As you can see on the topmost screenshot of this post, the place where the longest event takes place can get hidden behind the other overlapping event, which is fine, but again can create visual clutter as you may want to try and decipher the letter bits that peep out behind the front event. To reduce this, the event above could get a slight shadow on its edge, or have the text behind it slightly fade to transparency, so as to create a clearer separation between overlapping events, especially when they are of the same colour.
Here are two imperfect mockups for both of these solutions: 

Image

Image

larsbonczek:
Hi @Cwpute, thank you for the feedback. I have implemented your suggestion and I actually prefer it :)

image

I'm not sure how far the shadow should extend, though (or if it should only be drawn on top of other events and not on the background).

Cwpute:
Oh you actually implemented both ! This looks good yeah !
Now to see if that shadow translates well with other near events without a shadow.

larsbonczek:
I have made some more small changes and performance improvements (hopefully). This is what it looks like now:

Without staggered layout:
Screenshot from 2026-06-02 21-14-45

With staggered layout (light and dark theme):
Screenshot from 2026-06-02 21-12-24   Screenshot from 2026-06-02 21-12-39

Note that this will automatically fall back to the default column layout if there is not enough unobscured vertical space for any given event in a group. I think this is a good compromise since you can usually zoom in more to return to the staggered layout (except when two events start and end at the exact same time, I didn't come up with a solution for that).
You can see it in this screenshot where I zoomed out a bit (notice change at top right):
Screenshot from 2026-06-02 21-24-18

I will test this a bit more in the coming days and then submit the PR if I don't find any more problems/improvements.

@Gitsaibot could you please approve the workflow again so others can test this aswell?

larsbonczek:
I have done some final refactoring and squashed the commits on my branch. The feature is now ready for review.

Changes made:
- Added a new user preference "Staggered display" to General settings which enables the staggered layout
- Updated `DayView.remeasure` to fetch the value of the preference and pass it to `Event.computePositions`
- Updated `Event.computePositions` and `Event.doComputePositions` to compute the staggered layout based on the new parameter `drawStaggered`
  - Events are now grouped into sets of overlapping events first and then the positions are computed for each group of events separately by the new method `Event.computeGroupLayout`
  - The staggered layout code attempts to place events in as few staggered columns as possible, but since it's a greedy algorithm it might not find the optimal solution
  - If the start of an event is obscured by an event in a higher column, the staggered layout code computes by how much the event text needs to be moved down in order to be visible (`Event.setTextOffsetMillis`)
  - If at least one event in a group does not have enough contiguous unobscured vertical space to display its text (and receive tap events), the entire group falls back to the default column layout
- Updated `EventGeometry.computeEventRect` to compute the event rect for staggered events and to compute the text position from the precomputed text offset
- Updated `DayView.drawEvents` to draw events sorted by column if staggered display is enabled (to draw higher-column events on top of lower ones)
  - This also now passes the left and right bounds of the day being rendered to `drawEventRect`
- Updated `DayView.drawEventRect` to draw a shadow below staggered events (clipped to the bounds of the day being rendered)
- Refactored `DayView.setupTextRect` and `DayView.setupAllDayTextRect` and updated calling code
  - Both now initialize the text rect, which was previously done in `drawEventRect`
  - Both now return false if the rect is empty
  - `setupTextRect` was renamed to `getEventTextRect` and uses `event.textTop` to compute the text rect (which includes the precomputed offset)
  - `setupAllDayTextRect` was renamed to `getAllDayEventTextRect`
- Updated `DayView.findSelectedEvent` to sort selected events by column in descending order if staggered display is enabled to make sure the top-most events always receive the tap events

Please note:
- Like before, `Event.computePositions` only supports 64 columns (since it uses a 64-bit bitmask to track occupied columns). I have not fixed this, but added a TODO
- I only added the name and the description of the new setting in English. Should I add other languages as well?

